Issue #3, opened 2026-09-10
M1: reproducible Ouro benchmark and profiler baseline


hsliuustc0106:
## Precision requirements — 2026-09-12

For new work, **BF16 is the primary GPU inference, kernel-validation, quality, and performance target**. Use BF16 weights, ordinary activations, and KV storage; specify accumulation independently and retain FP32 in numerically sensitive reductions/intermediates. Full-model FP32 is an optional diagnostic/reference configuration, not the default prerequisite. See the [precision policy](https://github.com/hsliuustc0106/vllm-lt/blob/c9e646ace2d8198a52e522ea4d71129f1b143c58/docs/precision-policy.md).

BF16 profiling, matched optimization A/B, and quality data collection may proceed alongside the original Q1 reference diagnosis after each run's functional, finiteness, and ownership prerequisites pass. Report observations, reference fidelity, and qualification separately. Preserve exact cache/data-movement/RNG invariants and metadata-only A/B behavior; differing arithmetic paths use independently justified, predeclared BF16 numerical and decision/quality criteria. Universal bitwise or top-1 equality is not a blanket mixed-precision requirement, and near-tie exemptions cannot be invented after a run.

For dtype choices, follow explicit original Ouro paper guidance first, then the pinned release configuration and inference code. The [source audit](https://github.com/hsliuustc0106/vllm-lt/blob/c9e646ace2d8198a52e522ea4d71129f1b143c58/docs/paper-notes.md#original-ouro-paper-and-precision-guidance) distinguishes published FP32 operations from backend/project choices; the paper has no per-kernel inference recipe.

This amendment supersedes conflicting dtype/eligibility requirements below for successor work. All dated contracts, raw results, original failures, and historical acceptance decisions remain unchanged. New criteria need a separately frozen contract; no benchmark or qualification is claimed by this edit.

The historical M1 schema hard-codes FP32. Add a versioned BF16 contract/schema/runner path before new execution, preserving old reproduction. Keep 1,024 physical pages and report the BF16 pool as 3 GiB; the historical FP32 pool was 6 GiB.

Parent roadmap: #2. Milestone **M1**, the first implementation follow-up to preview #1.

## Problem and goal

The preview can execute real Ouro weights and has bounded correctness evidence, but we do not yet know whether its dominant costs are model computation, attention, host metadata, routing/readback, or KV propagation. The current checkpoint validation script is a correctness smoke, not a performance harness.

[...]

Issue #2, opened 2026-09-10
Roadmap: faster Ouro inference first


hsliuustc0106:
## Precision requirements — 2026-09-12

For new work, **BF16 is the primary GPU inference, kernel-validation, quality, and performance target**. Use BF16 weights, ordinary activations, and KV storage; specify accumulation independently and retain FP32 in numerically sensitive reductions/intermediates. Full-model FP32 is an optional diagnostic/reference configuration, not the default prerequisite. See the [precision policy](https://github.com/hsliuustc0106/vllm-lt/blob/c9e646ace2d8198a52e522ea4d71129f1b143c58/docs/precision-policy.md).

BF16 profiling, matched optimization A/B, and quality data collection may proceed alongside the original Q1 reference diagnosis after each run's functional, finiteness, and ownership prerequisites pass. Report observations, reference fidelity, and qualification separately. Preserve exact cache/data-movement/RNG invariants and metadata-only A/B behavior; differing arithmetic paths use independently justified, predeclared BF16 numerical and decision/quality criteria. Universal bitwise or top-1 equality is not a blanket mixed-precision requirement, and near-tie exemptions cannot be invented after a run.

For dtype choices, follow explicit original Ouro paper guidance first, then the pinned release configuration and inference code. The [source audit](https://github.com/hsliuustc0106/vllm-lt/blob/c9e646ace2d8198a52e522ea4d71129f1b143c58/docs/paper-notes.md#original-ouro-paper-and-precision-guidance) distinguishes published FP32 operations from backend/project choices; the paper has no per-kernel inference recipe.

This amendment supersedes conflicting dtype/eligibility requirements below for successor work. All dated contracts, raw results, original failures, and historical acceptance decisions remain unchanged. New criteria need a separately frozen contract; no benchmark or qualification is claimed by this edit.

Next precision work: implement versioned BF16 harness support, freeze the mixed-precision numerical/decision contract, and measure metadata reuse with matched BF16 A/B. Collect fixed-depth/adaptive BF16 quality evidence alongside the reference investigation. The historical FP32 M2 gain does not establish the BF16 gain.

Track the agreed priority: **faster Ouro-1.4B inference on one GPU**, with vLLM-style architecture, loop-level batching, and last-exited KV semantics.

[...]

Issue #8, opened 2026-09-10
Q2: screen Ouro task quality and establish a cached external baseline


hsliuustc0106:
## Precision requirements — 2026-09-12

For new work, **BF16 is the primary GPU inference, kernel-validation, quality, and performance target**. Use BF16 weights, ordinary activations, and KV storage; specify accumulation independently and retain FP32 in numerically sensitive reductions/intermediates. Full-model FP32 is an optional diagnostic/reference configuration, not the default prerequisite. See the [precision policy](https://github.com/hsliuustc0106/vllm-lt/blob/c9e646ace2d8198a52e522ea4d71129f1b143c58/docs/precision-policy.md).

BF16 profiling, matched optimization A/B, and quality data collection may proceed alongside the original Q1 reference diagnosis after each run's functional, finiteness, and ownership prerequisites pass. Report observations, reference fidelity, and qualification separately. Preserve exact cache/data-movement/RNG invariants and metadata-only A/B behavior; differing arithmetic paths use independently justified, predeclared BF16 numerical and decision/quality criteria. Universal bitwise or top-1 equality is not a blanket mixed-precision requirement, and near-tie exemptions cannot be invented after a run.

For dtype choices, follow explicit original Ouro paper guidance first, then the pinned release configuration and inference code. The [source audit](https://github.com/hsliuustc0106/vllm-lt/blob/c9e646ace2d8198a52e522ea4d71129f1b143c58/docs/paper-notes.md#original-ouro-paper-and-precision-guidance) distinguishes published FP32 operations from backend/project choices; the paper has no per-kernel inference recipe.

This amendment supersedes conflicting dtype/eligibility requirements below for successor work. All dated contracts, raw results, original failures, and historical acceptance decisions remain unchanged. New criteria need a separately frozen contract; no benchmark or qualification is claimed by this edit.

Use fixed-depth BF16 (B16) as the primary quality baseline and adaptive BF16 (A16) as its paired comparison. F32 is an optional, separately justified sensitivity control. New cached external timing uses matched BF16 with an explicit generation/replay-equivalence contract. Historical FP32 quality/timing results and their limitations remain intact.

Parent roadmap: #2. Depends on M1 #3 for measurement/reporting and [Q1 #4](https://github.com/hsliuustc0106/vllm-lt/issues/4) for numerical qualification. BF16 quality and external-baseline data collection can proceed alongside the historical Q1 diagnosis. Numerical and quality qualification remain separate decisions.

[...]

PR #10, opened 2026-09-10
Add reproducible Ouro inference benchmarks (M1)

hsliuustc0106:
The preview had correctness checks but no reproducible inference baseline. This adds CPU-only planning, real-weight replay/timing/profiling, and offline report validation for seven frozen Ouro FP32 workload cells. Production changes are limited to extracting the existing private exit predicate; replay preserves real gate, coda, sampling, and KV work. Implements M1 in #3 under roadmap #2.

The baseline completed all 7 feasibility, 7 warmup, 14 measured, and 2 profiler executions on a reserved NVIDIA L20X in 744.5 seconds. All 30 records passed offline validation, including both ABBA comparisons and request/KV cleanup. Final allocated/reserved GPU memory reached zero and the reservation released.

W4 mixed-depth replay measured **107.7–109.6 tokens/s with refill**, versus **84.1–86.2 without refill**: +27.1% and +28.1% in matched pairs. W5's uniform-depth control is inconclusive. These two observations per cell do not qualify BF16, adaptive quality, or performance against another engine.

Trace correlations attribute 6,144 W1 HtoD copies and matching synchronizations to repeated per-layer metadata, motivating M2 metadata reuse. The automatic report field gives a kernel-time ranking; the optimization proposal requires the separate trace analysis. W4 attribution accounts for interleaved prefill and benchmark page-scan overhead.

- [Reproduction guide](https://github.com/hsliuustc0106/vllm-lt/blob/codex-m1-benchmark/docs/benchmarks.md)
- [Baseline results and acceptance evidence](https://github.com/hsliuustc0106/vllm-lt/blob/codex-m1-benchmark/docs/benchmarks/m1-20260910.md)
- [Downloadable evidence](https://github.com/hsliuustc0106/vllm-lt/releases/tag/m1-evidence-20260910): all 96 raw report inputs, both full traces, original reports, and retained attribution scripts/results. The original producer of the older W1 analysis JSON is unavailable; its trace and reusable analyzer are included.

Measurements and the original report are frozen at `eb36183ecc980090310cdb08fbc38b2a371a0f61`. Later report fixes clarify ranking and CUDA stream-span semantics and explicitly exclude failed synchronization results; they add no measurements or speed claims.

Validation: the baseline passed 154 CPU tests, Ruff lint/format, and whitespace checks; 15 GPU tests remained opt-in. The report/runner follow-up passed 28 targeted CPU tests with CUDA discovery and initialization blocked. Documentation cleanup checked local links/anchors, shell syntax, CLI options, workload definitions, and all seven result rows against the retained JSON. It corrects one TPOT rounding typo and replaces the standalone review note with guide/report explanations; raw measurements are unchanged.


hsliuustc0106 (commented):
Review pass (code + tests + report cross-check; CPU suite re-run on L20X-8: 154 passed / 15 GPU-skipped, ruff check+format clean — matches the PR claim).

**Verdict: approve with comments.** The engine change is verifiably behavior-neutral (`_should_exit` extraction is the only production diff), replay fidelity is proven by tests that count gate/coda/sampling work rather than assert it, and token accounting is enforced three ways. Body claims reproduce from the committed report: +27.1%/+28.1% refill pairs (109.551/86.179, 107.701/84.056), 744.5 s, 30 records, 6,144 W1 HtoD copies, W5 inconclusive.

Minor items, none blocking:
- **[P3]** `vllm_lt/benchmarks/report.py:614-633` — `_next_action` nominates the largest GPU kernel group, but the published M2 nomination (metadata reuse) came from the out-of-band correlation scripts; the automated field alone does not reproduce the report's next target. Disclosed in the doc — consider wording the harness output as "kernel-time ranking" only.
- **[P3]** `vllm_lt/benchmarks/runner.py:199` — `begin.record()` precedes the enqueue loop, so `gpu_stream_span_ms` includes host submission time (documented in docs/benchmarks.md; a code comment would prevent misreading).
- **[P3]** `vllm_lt/benchmarks/report.py:301-307` — failed runs (`synchronized_ns=None`) fall into `summarize_records` unconditionally and record the less useful "must be an integer" error instead of the collector's clearer boundary-unavailable message. Cosmetic.

hsliuustc0106:
## Owner review (cross-implementation comparison with vllm-loop-serving)

Reviewed from the perspective of the sibling implementation at [vllm-loop-serving](https://github.com/hsliuustc0106/vllm-loop-serving) (serial reference + M0–M3 PRs #8–#11). Verdict: **approve-shaped; no production concerns.**

- The production diff (extracting `_should_exit`) is minimal and semantics-preserving. Good restraint — the benchmark package is the deliverable, and the engine change is exactly what replay needed and nothing more.
- **The replay methodology is the part worth keeping**: freezing token and exit histories while swapping schedulers, with gate/coda/sampling/KV work kept real, cleanly isolates scheduling effects from generation divergence. We measured scheduling numerics in vllm-loop-serving with teacher-forced trajectories but never built the replay-throughput harness; our M3 refill measurement will adopt this W1/W4/W5 cell design.
- W4 refill **+27.1/+28.1%** on matched pairs with the W5 uniform-depth control inconclusive is the first measured scheduling benefit of loop-level batching in either codebase, and the heterogeneity dependence is the expected signature. Worth stating on the roadmap as the working hypothesis for when refill pays.
- Side benefit of the extraction: it puts the `< 1.0` threshold guard (`params.exit_threshold < 1.0 and ...`) in one readable place. Porting note for symmetry: [vllm-loop-serving's serial exit policy](https://github.com/hsliuustc0106/vllm-loop-serving/blob/main/src/looped_serving/runtime.py) lacks this guard and has a latent early-exit bug under float saturation — remaining probability underflows to exactly 0.0 when sigmoid rounds to 1.0 (gate logits ≳ 17 in FP32), making `1 - remaining >= 1.0` fire at threshold 1.0. Your predicate is the correct semantics; we are porting it.

Issue #4, opened 2026-09-10
Q1: qualify BF16 numerics and validate last-exited KV independently


hsliuustc0106:
## Precision requirements — 2026-09-12

For new work, **BF16 is the primary GPU inference, kernel-validation, quality, and performance target**. Use BF16 weights, ordinary activations, and KV storage; specify accumulation independently and retain FP32 in numerically sensitive reductions/intermediates. Full-model FP32 is an optional diagnostic/reference configuration, not the default prerequisite. See the [precision policy](https://github.com/hsliuustc0106/vllm-lt/blob/c9e646ace2d8198a52e522ea4d71129f1b143c58/docs/precision-policy.md).

BF16 profiling, matched optimization A/B, and quality data collection may proceed alongside the original Q1 reference diagnosis after each run's functional, finiteness, and ownership prerequisites pass. Report observations, reference fidelity, and qualification separately. Preserve exact cache/data-movement/RNG invariants and metadata-only A/B behavior; differing arithmetic paths use independently justified, predeclared BF16 numerical and decision/quality criteria. Universal bitwise or top-1 equality is not a blanket mixed-precision requirement, and near-tie exemptions cannot be invented after a run.

For dtype choices, follow explicit original Ouro paper guidance first, then the pinned release configuration and inference code. The [source audit](https://github.com/hsliuustc0106/vllm-lt/blob/c9e646ace2d8198a52e522ea4d71129f1b143c58/docs/paper-notes.md#original-ouro-paper-and-precision-guidance) distinguishes published FP32 operations from backend/project choices; the paper has no per-kernel inference recipe.

This amendment supersedes conflicting dtype/eligibility requirements below for successor work. All dated contracts, raw results, original failures, and historical acceptance decisions remain unchanged. New criteria need a separately frozen contract; no benchmark or qualification is claimed by this edit.

The first-pass two-dtype case table and exact-top-1 gates below remain the historical diagnostic specification. A successor contract targets BF16, documents per-operation accumulation/casts, keeps the pinned official and independent LAST-EXITED checks, and uses FP32 only for a stated diagnostic question. Distinguish same-state local error from accumulated-state drift and report the old failures under their original policy. An eager reference rounding difference alone does not justify promoting every kernel to FP32.

[...]

PR #11, opened 2026-09-10
Default Ouro inference to BF16 and remove one-off experiment harnesses

hsliuustc0106:
Defaults the offline CLI, `LLM(...)`, and checkpoint loading to BF16, matching serving in #23. Explicit dtype overrides remain available; supplied model objects retain their dtype, and KV storage follows the model.

Removes the one-off M1/Q1 harnesses, contracts, fixtures, reports, validation adapters, copied reference code, and their dedicated tests. Retains the inference runtime, focused engine/model/attention/KV tests, and their dense reference. The README covers usage and tests; the design doc covers runtime semantics. Historical evidence remains in the [experiment archive](https://github.com/hsliuustc0106/vllm-lt/releases/tag/implementation-notes-archive-20260913).

Validation: 54 CPU tests passed; 15 GPU tests skipped. BF16/FP32 smoke checks, lint, formatting, imports, and packaging configuration passed. Documentation links, anchors, and the README CPU example passed. No new real-checkpoint accuracy or performance results.


hsliuustc0106 (commented):
Review pass (oracle independence, official-model fairness, fixture coverage, report tamper tests; CPU suite re-run on L20X-8: 353 passed / 26 GPU-skipped, ruff clean).

**Verdict: approve with comments.** The serial oracle is genuinely independent (imports only `OuroConfig` + torch; implements its own RoPE/RMSNorm/attention/gate; tests monkeypatch native helpers to fail), last-exited semantics are exactly tested (shallow-exit-then-deeper neighbor, block boundaries, hole rejection, NaN-poisoned storage, exact hazard CDF, threshold-1.0), the official comparison is dtype/normalization-fair, and the recorded BF16 decision (keep experimental, original tolerances retained, explicit counterexamples) is supported by the evidence.

Minor items:
- **[P3]** `vllm_lt/validation/runner.py:148-156` — oracle cleanup evidence is self-certifying: `requests_remaining`/`used_blocks` are computed after `oracle.close()` in the same `finally`, so they can only read 0/0 unless `close()` itself raised. The native path records real engine state; capture oracle state before close or assert `close()` raised.
- **[P3]** `serial_oracle.py` and `models/reference.py` encode the same reading of the official equations, so a shared misreading would evade both in-repo references; the 8 pinned-official executions are the only external cross-check (and in BF16 the official model itself disagrees with the oracle at L256-F0). Keep the official comparison mandatory in follow-ups — worth an explicit line in q1-validation.md.
- **[P3]** `vllm_lt/validation/native.py:27-52` — `snapshot_native` reads private `KVCacheManager` internals; acceptable for a validation adapter, note it as a refactor tripwire.

One intent check: `schema.py:464` tightens FP32 bounds to 0.001/1e-4 (historical 0.25/0.02) — stricter is fine and documented in q1-validation.md, just confirming it was deliberate.

hsliuustc0106:
### Numerical review: separate same-state arithmetic from accumulated-history error

Reviewed head `56b5456c23354193c38838781157e849956d2053`. Keeping the BF16 failure and official-reference discrepancy visible is the right disposition. The independent LAST-EXITED oracle is stronger than treating another serving engine as ground truth.

A useful next diagnostic is a three-way **same-state** replay at the first divergent boundary: snapshot identical token position, depth, hidden input, populated KV and probability state; execute serial BF16, batched BF16, and FP32 arithmetic over the same BF16 checkpoint. Then keep a separate **same-token-history, independent-state** experiment for accumulated error. Identical teacher token IDs alone do not imply identical hidden/KV inputs.

This distinction mattered in [our M3 diagnosis](https://github.com/hsliuustc0106/vllm-loop-serving/tree/32b4473/docs/validation-results/m3-norm-replay): a shape-dependent RMSNorm reduction changed BF16 rounding, and the batched output was closer to the FP64-rounded local reference. That is a concrete diagnostic example, not evidence that vllm-lt has the same cause.

For a separately versioned successor contract, compare each request/step's FP32 error to its own serial baseline; keep max and RMS checks separate for logits, per-depth gates, and each bank's historical/new-computed/new-fallback KV. Record top-two logit margins and distance to the exit threshold. Do not allow a long prefix or another gate depth to absorb a local regression.

Our [paired candidate](https://github.com/hsliuustc0106/vllm-loop-serving/blob/7658031/docs/paired-precision-contract.md) proposes `E_batch <= 1.25 * E_serial + 1e-6`, but **those numbers are not validated for this project and should not be imported as accepted tolerances**. Our fresh held-out run failed exact token agreement before numerical evaluation. Preserve this PR's original question and failures; no near-tie exemption or threshold widening should silently turn discrete failures into passes.

Suggested exit criteria for the next diagnostic: localize the first mismatch, establish whether the inputs are actually identical, preserve current-token fallback and prior-prefix invariants, and classify numerical-policy failure separately from semantic/cache failure. Q1 can then reach a supported negative diagnosis without implying BF16 deployment approval. Related issue #4.

hsliuustc0106:
## Owner review — keep draft; two suggestions toward closing AC-Q1-05

Preserving the unresolved BF16 diagnosis rather than qualifying is the right call. Cross-repo context: [vllm-loop-serving M2](https://github.com/hsliuustc0106/vllm-loop-serving/pull/10) hit the same wall independently — batched-vs-serial BF16 max logit diff 0.125 at atol/rtol 0.02, max 0.26 against an FP32 oracle, whole-history KV max 11.34 vs ~0.067 per newly computed token, identical discrete tokens and states, and a frozen held-out candidate that honestly failed. Combined with this PR's result (0.5625–0.625 at depth 4 against dense, and the official model itself disagreeing with the incremental BF16 reference in two bounded checks), the convergent reading is: **intrinsic depth-accumulated BF16 noise, not a caching or batching defect in either engine.** Two concrete suggestions:

1. **Slice the boundary observations by region.** In our M2, separating newly computed KV from carried historical-prefix error is what made the numbers interpretable (11.34 whole-history vs 0.067 per new token distinguishes "prefix noise carried forward" from "per-step growth"). Your per-layer/depth observations contain the same separability; cutting them new-vs-carried should classify the official-vs-incremental residual and possibly close AC-Q1-05.
2. **Add a mixed-precision arm** — FP32 residual stream/hidden + KV with BF16 (or FP16) projections only. Both repos jumped from BF16-everything to FP32-everything; the middle point would establish whether most of the error disappears at less than full FP32 cost, and it is one bounded experiment.

Also worth recording in the validation doc: two independent codebases — ours reusing the pinned official modeling code, yours a native port with paged KV — produce matching FP32-oracle agreement and identical BF16 token streams. That is mutual evidence the last-exited KV semantics are implemented correctly in both, independent of the dtype question.

PR #12, opened 2026-09-10
Reuse Ouro KV metadata across recurrent layers (M2)

hsliuustc0106:
Ouro previously rebuilt paged-KV metadata in every physical layer. This change prepares one borrowed descriptor per recurrent traversal and reuses its positions, write addresses, block tables and context lengths across the layers. Allocation identity and initialized-history checks remain enforced; cache layout, sampling, gates and public cache adapters preserve their behavior.

Based on `codex-m1-benchmark`, where PR #11 was merged. Its BF16 default and experiment-harness deletions are preserved. The remaining diff contains the cache/model implementation, focused prepared-KV tests, and the precision policy.

## Validation

CPU checks: **84 passed, 15 GPU-only tests skipped**, with CUDA discovery and initialization blocked. Ruff lint/format, documentation links/anchors and diff checks pass. Focused tests cover descriptor lifetime, stale allocation rejection, initialized-prefix checks, duplicate-write rejection, and one metadata preparation per traversal.

A separate BF16 GPU comparison on 2026-09-13 followed PR #25's A/B protocol, comparing immediate base **`f97a37e` (A)** with PR head **`886c0d2` (B)**:

- GSM8K-87: **59/87 (67.82%) on both sides**, with zero paired correctness losses or generated-token disagreements. This is a selected regression cohort, not a held-out quality estimate.
- All **257 captured BF16 tensor boundaries**, including populated KV state, matched exactly in the selected numerical case. Two seeded stochastic smoke requests also matched tokens, exit depths and final request RNG states. These checks do not establish general adaptive task quality.

## Engine performance

Speed was measured using a **custom external experiment driver** (`worker.py`, orchestrated by `controller.py`) calling `LLMEngine.add_request()` and `engine.step()` directly. It is not a checked-in repository benchmark or `vllm bench serve`. Only the shared GSM8K accuracy evaluator was copied from repository commit `eb0ee74`.

Both sides used the same NVIDIA L20X GPU, CPUs 24–31 on NUMA node 0, pinned Ouro-1.4B checkpoint, BF16 parameters/KV, Triton attention, and fixed four-loop greedy decoding with EOS ignored. Scheduler settings were `mode="refill"`, **`max_num_seqs=8`**, and **`max_num_batched_tokens=128`**, with a 512-page KV pool (16 tokens/page). Actual request concurrency was 1 or 8 as shown below.

[...]

PR #13, opened 2026-09-10
feat: support inactive Ouro recurrent rows

### LINKED ISSUE #6: M3: validate and capture one recurrent traversal with CUDA graphs
## Precision requirements — 2026-09-12

For new work, **BF16 is the primary GPU inference, kernel-validation, quality, and performance target**. Use BF16 weights, ordinary activations, and KV storage; specify accumulation independently and retain FP32 in numerically sensitive reductions/intermediates. Full-model FP32 is an optional diagnostic/reference configuration, not the default prerequisite. See the [precision policy](https://github.com/hsliuustc0106/vllm-lt/blob/c9e646ace2d8198a52e522ea4d71129f1b143c58/docs/precision-policy.md).

BF16 profiling, matched optimization A/B, and quality data collection may proceed alongside the original Q1 reference diagnosis after each run's functional, finiteness, and ownership prerequisites pass. Report observations, reference fidelity, and qualification separately. Preserve exact cache/data-movement/RNG invariants and metadata-only A/B behavior; differing arithmetic paths use independently justified, predeclared BF16 numerical and decision/quality criteria. Universal bitwise or top-1 equality is not a blanket mixed-precision requirement, and near-tie exemptions cannot be invented after a run.

For dtype choices, follow explicit original Ouro paper guidance first, then the pinned release configuration and inference code. The [source audit](https://github.com/hsliuustc0106/vllm-lt/blob/c9e646ace2d8198a52e522ea4d71129f1b143c58/docs/paper-notes.md#original-ouro-paper-and-precision-guidance) distinguishes published FP32 operations from backend/project choices; the paper has no per-kernel inference recipe.

This amendment supersedes conflicting dtype/eligibility requirements below for successor work. All dated contracts, raw results, original failures, and historical acceptance decisions remain unchanged. New criteria need a separately frozen contract; no benchmark or qualification is claimed by this edit.

New inactive-row, persistent-buffer, and eager/graph comparisons target BF16 with the same declared accumulation policy on both sides. Keep exact mask/ownership/lifetime invariants. A historical Q1 diagnosis is not a prerequisite to collecting these A/B results; promotion still requires the new contract's numerical and behavioral gates.

Parent roadmap: #2. Milestone **M3 — one recurrent traversal under CUDA graphs**.

Dependencies: M1 profiler/benchmark #3, an accepted tensor/metadata boundary from [M2 #5](https://github.com/hsliuustc0106/vllm-lt/issues/5) or an equivalent narrow extraction, and the selected dtype contract from [Q1 #4](https://github.com/hsliuustc0106/vllm-lt/issues/4). BF16 is the primary target, with accumulation precision fixed across A/B; Q1 reference diagnosis and result collection can proceed together.

## Problem and decision gate

The current runner launches one recurrent traversal eagerly and reads the actual gate probabilities before CPU routing. Graph replay may reduce launch overhead, but #3 must first establish that this cost matters. Attention work may take priority. Start with a recorded hypothesis, bounded target shapes, memory/capture budgets, primary inference metric, and maximum control-workload regressions. A graph implementation without representative benefit is not a performance result.

The present attention kernel has no safe inactive-row contract: zero context length executes no softmax loop and can store `0/0`. Cache writes also assume valid rows. Padding current inputs and calling capture is therefore unsafe.

## Goals, constraints, and non-goals

Capture exactly one Ouro shared-core traversal, including its end-of-loop normalization and actual gate tensor. Retain synchronous host routing and LAST-EXITED finalization. Preserve positions while depth advances, full-depth prefill, full-lifetime reservation, request isolation, cancellation, and eager numerical behavior.

Exclude lookahead gates, asynchronous scheduling, prefill/coda capture, stochastic sampling capture, allocator redesign, attention tuning, and changed dtype policy. Keep unsupported inputs on an explicit eager path. This milestone should use separate reviewable PRs for inactive-row correctness, persistent execution buffers, and capture/replay; correctness prerequisites need no speed claim.

## 1. Explicit inactive rows

Extend the prepared tensor contract with a live-row mask. Illustrative internal fields:

```python
@dataclass
class RecurrentInputs:
    hidden: torch.Tensor          # [bucket_rows, hidden_size]
    positions: torch.Tensor       # [bucket_rows]
    block_tables: torch.Tensor    # [bucket_rows, table_width]
    lengths: torch.Tensor         # [bucket_rows]
    write_blocks: torch.Tensor    # [bucket_rows]
    write_offsets: torch.Tensor   # [bucket_rows]
    active: torch.Tensor          # [bucket_rows], bool
```

Every inactive row has a defined finite attention output, performs no KV write/read, and cannot update request state or routing. Mask address loads and memory operations before an invalid padded address can be dereferenced. Merely multiplying a completed result by zero cannot repair a prior invalid access or NaN. Do not use a real request's pages as dummy padding.

For active rows, preserve existing causal lengths and their independently selected depth tables. The first implementation can retain a packed live-row prefix while testing the mask itself with inactive rows between active ones. Support zero live rows as a safe no-work case; production dispatch need not replay an empty batch. Keep padded rows out of coda and RNG handling.

## 2. Stable storage and state ownership

The runner owns persistent input/output buffers for a small declared bucket set. Buffer addresses and shapes used by capture remain stable. Copy current metadata and gather request hidden states into those buffers before replay. Allocation/address validation and request lookup remain on the host, outside the captured function.

```python
def select_bucket(live_rows: int, required_table_width: int) -> BucketKey | None: ...
def prepare_replay(batch: SchedulerOutput, bucket: BucketKey) -> None: ...
def recurrent_tensor(inputs: RecurrentInputs) -> tuple[torch.Tensor, torch.Tensor]: ...
def publish_outputs(batch: SchedulerOutput, hidden: torch.Tensor) -> None: ...
```

Keep request state independent of reusable graph output storage. Assigning `request.hidden_state = graph_output[row]` would retain a view that later replay overwrites, including for a request waiting in coda. Copy live outputs into request-owned storage outside capture, and complete/order that copy before reusing graph buffers. This can retain the current request fields without adding a second scheduling slot system.

Free/reuse pages only after all task-owned operations referencing them complete. Use the existing ordered execution stream and explicit synchronization/event ownership where needed; do not introduce overlapping replay. Invalid execution aborts affected requests and prevents unsafe resource reuse. Never recycle pages following an unconfirmed device failure merely because a host callback ran.

Per-layer written-prefix bookkeeping currently happens in Python. For the tensor traversal, validate required preceding history before dispatch, preserve write-before-attend ordering inside every layer, and commit successful current-position bookkeeping before engine finalization/routing. The graph must not capture or replay Python bookkeeping side effects. Tests must show equivalent initialized-history behavior and failure invalidation.

## 3. Bounded capture and dispatch

Use a private runner configuration, initially opt-in, with an explicit ordered list of row/table-width buckets and byte/setup-time limits. It may be passed through existing engine construction plumbing if necessary; do not add a public graph plugin interface. A bucket key covers static properties affecting capture, including device, dtype and fixed tensor shapes. Metadata values such as position, depth-selected page addresses, and valid lengths remain dynamic tensor contents.

At setup, warm and capture against task-owned scratch state with initialized history. Never warm up or capture by modifying a user's live request. If scratch pages come from the same fixed pool, account for capacity, restore ownership safely, and verify later replay reads current metadata rather than baked-in scratch addresses. Initialize inactive buffers explicitly.

Capture the tensor core only: physical layers, shared normalization, and gate output. After replay, perform the actual gate probability conversion/readback and the existing CPU update. Preserve cumulative hazards before minimum exit depth. LAST-EXITED propagation stays outside capture and precedes dependent future execution. This is launch optimization, not the paper's lookahead method.

Fall back eagerly for unconfigured shapes or insufficient capture budget, recording the reason and invocation count. Do not silently retry a device/correctness failure as an eager fallback. Retain a bounded graph cache; no unbounded capture for newly observed lengths. Capture setup, graph-resident memory, coverage, and amortization are separate report fields.

## Validation and experiment contract

CUDA-free tests cover bucket selection, descriptor ownership, live-row bookkeeping, fallback decisions, and plan limits. Reserved kernel tests cover zero/one live row, every bucket boundary, invalid padding sentinels, fragmented pages, mixed depths, block crossings, and finite inactive outputs. Verify no padded-row KV changes, including neighboring request slots.

Compare eager and replay hidden states/gates and complete token/depth histories using the retained numerical contract. Exercise refill/no-refill, requests finishing at different times, coda waits, cancellation/reuse, alternating buckets, and eager/graph transitions. Ensure capture scratch state cannot contaminate live history. Check initialized-prefix invariants and zero retained request blocks after cleanup.

For performance, reuse #3 timing and replay semantics, freezing exact devices, hardware/software, model, dtype/arithmetic, fixtures, cache policy and residency. Predeclare the independent variable and regression/memory/setup limits. Run a CPU probe, one excluded feasibility comparison, and the declared warmups and two measured runs per configuration unless another budget is recorded beforehand. All GPU work requires the verified exact-ID reservation and task-owned cleanup.

Include input preparation, output publication, gate transfer, routing and eager fallbacks in the end-to-end comparison. Separate capture/warmup cost from steady execution and report both observations/ranges. Stop at budget on inconclusive benefit, correctness failure, memory excess or unacceptable regressions. Do not grow buckets or workloads silently to find a favorable result.

## Acceptance criteria / 验收标准

Performance targets are filled from the M1 baseline before candidate measurements: `P > 0` is required improvement, control regression limits are numeric and nonnegative, and resource/setup limits include units. Throughput improvement is `100 * (candidate / baseline - 1)`; latency improvement is `100 * (1 - candidate / baseline)`. Apply the selected definition to each matched pair and the corresponding control observations. A TBD target or a target chosen after seeing candidate results cannot pass acceptance.

Each checked AC links a reproducible test/result artifact and implementation SHA. AC-M3-02 through AC-M3-05 may accept separate correctness-prerequisite PRs; they cannot close this performance milestone. Completion requires all criteria, including end-to-end benefit. BF16 tests use a predeclared mixed-precision contract linked to #4; an open historical diagnosis does not block data collection, and promotion requires the successor contract's gates.

- [ ] **AC-M3-01 — Measured rationale and frozen limits.** Link #3 launch-overhead evidence and record the tensor/metadata prerequisite from #5 or its equivalent. Before candidate measurements, freeze base/candidate SHAs and modifications, dtype/tolerances, bucket shapes, workload/control matrix, numeric primary improvement target `P%`, per-control regression limits, graph-memory/setup budgets, comparison order and variation rule. **Pass:** every limit is concrete, predates candidate measurements, and controls differ only by the declared graph change and recorded necessary consequences.

- [ ] **AC-M3-02 — Inactive rows cannot touch live state.** Reserved tests exercise zero/one live row, inactive rows between active rows, each configured row/table-width boundary, invalid padded addresses, fragmented pages and mixed depths. Poison neighboring KV slots and run a device memory-access checker where supported; otherwise disclose that limitation and retain guard/instrumentation evidence. **Pass:** inactive attention outputs are finite and defined; no invalid-address error, sentinel change, request-state change or RNG advancement occurs; active outputs match eager tolerances. Post-hoc multiplication by zero does not satisfy masked-access evidence.

- [ ] **AC-M3-03 — Stable storage and independent request lifetime.** Record captured tensor pointers across repeated replay and alternating buckets. Pause requests in coda while other requests replay; cancel and reuse request/pages; transition between eager and graph execution. **Pass:** captured addresses/shapes remain stable; paused hidden states remain unchanged until their own execution; output publication cannot alias reusable graph storage; all page reuse follows completion of its prior consumers. Successful cleanup leaves zero active requests and request-reserved pages.

- [ ] **AC-M3-04 — Correct traversal and bookkeeping.** Compare per-loop hidden states, gates, full populated KV prefixes and final token/depth histories under both scheduling modes. Include block boundaries, heterogeneous exits and injected pre-dispatch/update failures. **Pass:** all frozen numerical/discrete criteria hold; preceding history is checked before execution; every layer writes before attending; Python initialized-prefix bookkeeping advances only for valid execution before finalization. Capture contains no host routing/bookkeeping dependency, and actual cumulative gate readback retains the synchronous contract. Unconfirmed device failures cannot trigger unsafe page recycling.

- [ ] **AC-M3-05 — Bounded setup and explicit fallback.** Test scratch-state capture, changed runtime page addresses/positions, unsupported shapes, memory/setup limits and capture/device errors. **Pass:** live request state is untouched by setup; replay consumes current metadata; graph count/residency stays bounded; unsupported inputs take counted eager fallbacks with reasons. Device/correctness failures surface rather than silently retrying. Required CPU/kernel checks and lint/format pass; unavailable checks remain explicit gaps.

- [ ] **AC-M3-06 — Complete end-to-end performance gate.** After the CPU probe and excluded feasibility/warmup, run frozen A–B/B–A pairs on identical exact devices through the verified reservation. Timing includes per-step input preparation, output publication, actual gate transfer, routing and fallbacks; capture/setup is separately reported. **Pass:** both pairs meet `P%` for every declared primary target; every control observation remains within its regression limit; measured memory/setup stays within budget. Missing controls or averaging away a failed pair/control cannot pass. Benefit within observed variation under the frozen rule is **inconclusive** at the run limit, not grounds for extra unbudgeted trials.

- [ ] **AC-M3-07 — Evidence and adoption decision.** Link raw runs, manifests, individual values/ranges, capture costs, memory, hit/fallback counts, cleanup logs and AC outcomes. **Pass:** all gates support the stated optional/default adoption scope. Correct masking or capture alone is insufficient. Deferral/no-go keeps the milestone open unless explicitly closed **not planned**, with unmet gates visible; it never marks the optimization accepted.

Source snapshot: [current recurrent execution](https://github.com/hsliuustc0106/vllm-lt/blob/fa7a2ca5b833f2dd0f3e8db53ce111a7406496f7/vllm_lt/worker/model_runner.py), [attention padding behavior](https://github.com/hsliuustc0106/vllm-lt/blob/fa7a2ca5b833f2dd0f3e8db53ce111a7406496f7/vllm_lt/kernels/triton_attention.py), [written-prefix ownership](https://github.com/hsliuustc0106/vllm-lt/blob/fa7a2ca5b833f2dd0f3e8db53ce111a7406496f7/vllm_lt/core/kv_cache_manager.py).

hsliuustc0106:
Padding recurrent batches could dereference invalid KV addresses and produce NaNs for inactive rows. This adds an explicit live-to-physical row map, masked attention/KV scatter, and private padded eager execution that publishes only live outputs into independent storage. Production generation remains compact.

Based on `main` after integration PR #26 (`8d28ed7`), which delivered #11 and #12. The base merge preserves BF16 defaults, metadata reuse, the serving frontend and scheduler fairness changes, the GSM8K evaluator, and removal of the retired milestone harnesses. Runtime conflict resolution preserves this PR's existing inactive-row cache/model implementation; the diff against main is limited to 10 files (929 additions, 18 deletions). Dependent M3 experiment modules, fixtures and harness tests are also removed; the core inactive-row regression tests remain. The mixed-depth transition test now compares padded execution with compact execution using separate KV caches, rather than importing the removed serial oracle.

### Validation

- Current CPU suite: **197 passed, 49 GPU tests skipped**, with CUDA discovery, initialization and synchronization blocked. Ruff lint/format, local documentation links/anchors and diff checks pass. No new GPU execution was performed for the merge.
- Core tests cover poisoned inactive inputs, invalid addresses, MHA/GQA and head dimensions through 256, KV isolation, live-state publication, per-request RNG isolation, and compact-versus-padded execution.
- Historical reserved kernel review: **62 passed**, including 49 device cases, with final task-owned CUDA memory 0/0. This is prior evidence, not a rerun of the merged source.
- Historical real-checkpoint FP32 A/B: **65/65 executions, 52/52 kernel evaluations and 28/28 comparison streams**. The declared eager AC-M3-02 matrix passed. M3 remains open for persistent storage, capture/replay and performance; no BF16 model qualification or speed claim is made. No memory-access checker was available for that run.

The [frozen review record](https://github.com/hsliuustc0106/vllm-lt/blob/8ecd1feecacf019feae1c02c12e32dac1bf840a8/docs/reviews/pr13.md), [interface design](https://github.com/hsliuustc0106/vllm-lt/blob/8ecd1feecacf019feae1c02c12e32dac1bf840a8/docs/m3-inactive-rows.md), [acceptance report](https://github.com/hsliuustc0106/vllm-lt/blob/8ecd1feecacf019feae1c02c12e32dac1bf840a8/docs/benchmarks/m3-inactive-20260911.md) and [evidence release](https://github.com/hsliuustc0106/vllm-lt/releases/tag/m3-inactive-evidence-20260911) preserve the removed experiment tooling and historical results. Reproduction uses the frozen sources; this PR does not close #6.


hsliuustc0106 (commented):
Review pass on the kernel/masking changes. Remote verification on L20X-8: CPU suite 546 passed / 27 GPU-skipped, ruff clean; and via `gpu run` (reserved GPU 1): `tests/test_attention.py tests/test_inactive_kernels.py --run-gpu` → **29 passed**, which executes the new Triton masked path incl. GQA and dim buckets on device.

**Verdict: approve with comments.** The inactive-row math is sound: masked lengths/queries give an empty loop and `acc / tl.where(enabled, normalizer, 1.0)` yields exact +0.0 (signbit-checked); KV scatter never dereferences -1/2^30 sentinels; poisoned NaN hiddens are sanitized pre-projection with bias-safe zeroed gate logits; padded path gathers via `index_select` (no scratch aliasing) and never touches coda, per-request generators, or last-exited finalization. Tests assert `isfinite(...).all()` and whole-cache `torch.equal`, so NaN or a stray KV write would be caught.

Minor items:
- **[P3]** `vllm_lt/validation/m3_inactive_kernels.py:50-55` — the recorded validation matrix covers masked execution only at head_dim=128 with query_heads==kv_heads (MHA); the GQA / other-bucket masked combinations are exercised by the test suite (now confirmed on GPU) but not by the recorded validation evidence. head_dim 256 remains untested repo-wide.
- **[P3]** `vllm_lt/kernels/triton_attention.py:97-101` — an *enabled* row with length 0 yields 0/0 NaN (normalizer stays 0). The manager currently guarantees active length ≥ 1, so this is contract-only; a `tl.device_assert` or comment would harden it against future callers.
- **[P3]** `vllm_lt/kernels/triton_kv_write.py:11-31` — all strides are `tl.constexpr`, so each distinct cache/K layout triggers recompilation. Fine here; worth remembering when reused from a captured graph (layouts are then fixed per bucket, which bounds it).
- **[P3]** `vllm_lt/core/kv_cache_manager.py:390-396` — padded torch-backend `store` indexes `write_blocks` with a Python list (implicit per-layer H2D) and `attend` calls `active.tolist()` (per-layer sync). Acceptable for the private eager hook; should not migrate into the production path.

hsliuustc0106:
## Owner review — approve; the masking details are the good part

Inactive-row handling is where padded execution usually silently breaks, and the details here are all correct:

- `masked_fill(0.0)` instead of multiplication — poison NaNs in inactive inputs survive multiplication by zero-payload masks; this is the right call and the inline comment carries the rationale.
- Explicit gate-logit masking, with the bias rationale (zero hidden alone does not define padding — the gate has a bias). Easy to miss.
- The torch path gathers live rows before advanced indexing so `-1` sentinels never reach an indexer (`-1` under advanced indexing would alias the last block).
- `accumulator / tl.where(enabled, normalizer, 1.0)` avoids 0/0 at zero context length, and `HAS_ACTIVE` as constexpr keeps the compact path mask-free.

Design note for the record: vllm-loop-serving's batched runner took the opposite strategy for mixed-length rows — group rows by equal valid history length, no masks, at the cost of fragmented attention launches. The masked-row approach here keeps one launch per layer at the cost of wasted rows, which is the right tradeoff heading into static-shape graph capture. Worth recording in the M3 design doc so the choice isn't relitigated later.

Residual overhead candidate (not blocking): `written` bookkeeping is still per-row-per-layer Python (`allocation.written[depth][layer].add(position)` — 192 calls per 8-row traversal). If a later milestone targets host time again, folding the written-position update into the descriptor lifecycle is the next obvious consolidation.

hsliuustc0106:
Superseded by #27, a single commit on current main with the same source tree as updated head 25628b9467d341d269f6df926d93c5080dcdb9ef. Reserved GPU regressions, BF16 compact-versus-padded correctness, and compact-path performance will be rerun at #27’s exact head before merge. The new evidence will also be posted here; prior #12 measurements do not qualify this change.

hsliuustc0106:
GPU qualification at replacement #27 head `6a0cb35aafcd84a7f0334aac8f0323c514acf24d` **failed the frozen BF16 padded numerical gate**. Its source tree exactly matches updated #13 head `25628b9`. #13 is closed in favor of #27; #27 remains a draft and has not been merged.

- CPU: **197 passed, 49 GPU skipped**. Reserved kernel regression: **62 passed** (49 device + 13 CPU cases).
- Main/candidate compact execution: **1,027 tensor boundaries / 52,465,722 elements bitwise equal**. Padded final-logit bounds and all 25 held-history argmax rows pass; 24 actual live generated tokens, exits and RNG state per mode match. All values finite, all unused-cache guards preserved.
- Padded mixed-length BF16 execution exceeds the fixed hidden/gate bounds at **three elements in three boundaries**. See the raw values and allowed errors in the report. Bounds were not relaxed; no failed run was replaced. This establishes a numerical-contract failure, not by itself a memory-safety defect or task-quality loss.
- The stop rule prevented ABBA measured speed runs. Excluded feasibility timings do **not** qualify compact-path performance. Earlier #12 evidence is not used as qualification.
- All three workers finished with **0 allocated / 0 reserved CUDA bytes**; GPU3 was automatically released and verified AVAILABLE at 0 MB.

[Full report, protocol, raw tensors/logs, source archives and hashes](https://github.com/hsliuustc0106/vllm-lt/releases/tag/pr27-gpu-qualification-20260913). Archive SHA-256: `796aba6baa0e8c254e491b1b2d210beed72a32243a7dc8d6aabfab842e2b8fb5`.

#14 has not been advanced: its prerequisite has not passed and merged. Read-only inspection also found that its reusable-buffer setup currently rejects BF16; this needs attention before validating it against current main's default.

PR #14, opened 2026-09-10
feat: reuse eager Ouro decode buffers

hsliuustc0106:
Adds opt-in persistent eager decode buffers with generation and stream ownership, independent published request state, compact fallback for oversized batches and cache quarantine after uncertain completion. Stacked on #13.

Prior validation: 667 CPU tests passed; 16 GPU cases skipped. All 17 allocating-versus-persistent validation executions passed. This does not establish a speedup or hardware-fault recovery.


hsliuustc0106 (commented):
Review pass on lifecycle/reuse semantics. Remote verification on L20X-8: CPU suite 663 passed / 27 GPU-skipped, ruff clean.

**Verdict: approve with comments.** The PR claims only storage reuse with stable pointers and isolation — explicitly no speedup — and the evidence supports exactly that. Correctness checks out: slot mapping `2*i+1` is unique per traversal (max index 7 < 8); staging is fully re-initialized every call with sentinel fills; `hidden_in` is zeroed before `index_copy_`; published outputs are `index_select` copies with no alias to fixed buffers (asserted via storage_ptr); per-request RNG and paused-coda state survive buffer churn; lease/generation guards reject stale, superseded, reentrant, and reallocated-ID descriptors; cancellation → free → reuse is covered by fresh written-prefix bookkeeping, and equivalence is tested bitwise (atol=0) plus full-trajectory equality in both scheduler modes.

Minor items, all non-blocking:
- **[P3]** `vllm_lt/worker/model_runner.py:177-181` — with persistent enabled, an exception from *any* stage (including prefill, which never touches persistent storage) permanently disables the executor via `_settle_persistent_failure`. Conservative and tested, but overbroad.
- **[P3]** `vllm_lt/engine/llm_engine.py:84-96` — KeyboardInterrupt/SystemExit raised in `_update` after successful gate readback now abort scheduled requests and permanently disable the executor; the compact path propagates interrupts without cleanup. Documented, but a real interrupt-semantics change for the opt-in path.
- **[P3]** `vllm_lt/worker/model_runner.py:254` — nothing ties `SchedulerConfig.max_num_seqs` to the 4-live-row/32-column capacity; an engine with `max_num_seqs > 4` silently falls back to the compact path (counted in `fallback_counts`, never surfaced). Worth an assertion or warning before #15 builds capture on top.
- **[P3]** `vllm_lt/worker/model_runner.py:45` — the 256-KiB payload-cap formula hardcodes metadata sizes (1256+32), duplicating `_allocate_metadata_storage` specs; a shared constant would prevent drift (snapshot test currently catches it).
- **[P3]** `vllm_lt/engine/llm_engine.py:71` — under quarantine (unconfirmed stream completion), `abort_request`/`free` raise, so remaining requests/pages can never be reclaimed. Deliberate per the design doc, but callers get no drain path.

hsliuustc0106:
### Cross-engine design review: retain traversal leases and independent publication

Reviewed head `630a8fdc0dd47b6da68a3d30db6b851fc08af5c5`, including the #12 metadata and #13 inactive-row prerequisites. These are the most useful components to borrow into `vllm-loop-serving`: metadata is prepared per traversal, inactive rows cannot address live pages, and persistent storage has explicit allocation/generation ownership.

The key contract to preserve in future optimization is **borrowed traversal state versus independently owned request state**. Stable pointers alone are insufficient: a coda-held request must survive the next buffer refill, and a reused request ID/page must not validate an old prepared descriptor. The current generation/allocation checks and independent live-output publication address this; please keep these as prerequisites for any future copy elimination.

For #6's next phase, carry these tests through graph replay and real engine cancellation/slot reuse, including a request paused at CODA while another advances multiple loops. Distinguish a completed gate readback from later LAST-EXITED finalization; pages must remain owned until the latter's writes and consumers are settled. The current synchronous lifetime contract is useful and does not require implementing asynchronous lookahead now.

This is a positive design assessment, not a new performance finding or blanket approval of the stack. The published A/B isolates persistent versus allocating **padded** execution; retaining the opt-in scope and making no compact-path speed claim is appropriate. Our targeted CPU review suite, including persistent-buffer and KV ownership tests, passed at the downstream #15 snapshot; we did not rerun device lifecycle tests.

hsliuustc0106:
## Owner review — approve with concerns; the fixed bucket is the structural issue

Correctness and lifecycle machinery look sound (leases, generations, quarantine with preserved allocation identities, ownership-before-recycle, counted fallbacks). Two concerns:

1. **Hardcoded 4-live/8-physical/32-column capacity vs `SchedulerConfig.max_num_seqs = 8`.** `_prepare_into` raises past 4 live rows, so valid >4-request decode batches take the compact allocating fallback. With full-depth FP32 reservation at ~1.5 MiB/token, a 48 GB L20X still admits roughly 30 concurrent requests, so >4-row decode batches are the norm — the persistent path serves at most half the advertised concurrency. This directly explains the W4 fallback-only profile window observed in #15.
2. **Recommend a capture-size ladder instead of one fixed bucket + fallback**: capture 4/8/16 live-row shapes at config time (vLLM's own pattern — a small set of padded capture sizes sharing one graph pool). Same correctness, materially less bookkeeping than lease/generation/quarantine. If the ladder lands before #15's rework, the graph buckets and the persistent buckets should be the *same* ladder, not two parallel shape systems.

Smaller notes: the engine-side `BaseException` settlement choreography in `step()` is the highest-complexity-per-line region of this PR; consider extracting it into the runner (which already owns `_settle_persistent_failure`) so the engine loop stays readable. And `_prepare_into`'s six synchronous staging copies per decode step are fine, but the M2 accounting doc should note that the budget now reads "5 per traversal + 6 per decode step" so it doesn't drift.

PR #15, opened 2026-09-10
feat: capture bounded Ouro recurrent CUDA graphs

hsliuustc0106:
Adds opt-in FP32 recurrent CUDA graphs with scheduler-sized buckets, a shared graph pool, reusable capture stream and bounded setup/cleanup. Compact execution remains the default; routing and gate readback remain synchronous. Stacked on #14.

Prior validation: 798 CPU tests passed; 49 GPU cases skipped, followed by 38 targeted checks. Current-source GPU qualification and profiling/resource gates remain incomplete; CPU tests do not qualify replay or a performance improvement.


hsliuustc0106 (commented):
Review pass on the capture/replay mechanics. Remote verification on L20X-8: CPU suite 915 passed / 27 GPU-skipped, ruff clean; `tests/test_attention.py tests/test_inactive_kernels.py --run-gpu` at this head via `gpu run`: **29 passed**.

**Verdict: approve with comments — keep draft until the owned-pool GPU qualification passes.** The core graph-safety properties verify correct in code: static buffers are the persistent ones (`_prepare_into` copies in place; `_check_bucket` re-checks data_ptr/shape/stride on every dispatch); `hidden_in` is re-copied before each replay; outputs are published via fresh `index_select` tensors (no aliasing, test-asserted); gate logits land in `gate_out` inside the graph with host sigmoid+readback strictly before `_commit_decode_traversal` (order test-asserted); sigmoid/sampling stay outside; prefill stays on the compact path; buckets 1-2→4, 3-4→8, >4 rows or >32 columns or non-triton → counted compact fallback (no off-by-one found; 512-token boundary tested); per-bucket `MemPool` ownership with sync→reset→drop-outputs→drop-owners-last teardown and retention on failed close.

Findings:
- **[P2]** `vllm_lt/worker/model_runner.py:109,120` — `error.add_note(...)` requires Python 3.11+, but pyproject declares `requires-python >= 3.10`. On 3.10 the budget-decline refusal path raises `AttributeError` and masks the original `CaptureBudgetExceeded`. Gate the note or drop it (same pattern exists in some harness-only modules, lower stakes there).
- **[P2]** `vllm_lt/worker/recurrent_graph.py:696-699` — a second `settle_failure` after a recorded failure returns the cached `completion_confirmed` and silently drops the new error instead of appending to `failure["secondary"]` (only `record_cleanup_failure` appends). Observability gap on cascaded failures.
- **[P3]** `vllm_lt/worker/recurrent_graph.py:398-399` — `_restore_scratch` bypasses manager APIs (`del cache._allocations[...]`, `cache._free_blocks[:] = ...`). Private-by-design and hash-recorded, but it will diverge silently if release logic gains side effects.
- **[P3]** `vllm_lt/worker/recurrent_graph.py:513-528` — setup memory deltas use device-global `torch.cuda.memory_*` counters, not per-pool counters; fine for the global gates and honestly disclosed in the doc as not proving physical reclaim.
- **[P3]** CPU tests fake CUDA (fake replay re-runs the eager body), so they verify transactions/lifetimes, not real replay; the real-replay evidence is the recorded GPU correctness gate (15 model cases, 120 eager + 120 actual replay dispatches, 64 counted fallbacks, original tolerances) — that run passed, but no passing end-to-end GPU qualification exists yet for this candidate.

Context note: the stopped-attempt record (memory gate 532 vs 512 MiB, controller OSError, corrected reporter) is honest and the owned-pool retry (0e4516a) is the right fix path; a reservation for that qualification was running on GPU 0 of L20X-8 during this review and has since finished — assessment of its results belongs to that track.

hsliuustc0106:
### Graph review: distinguish graph mechanism from serving adoption

Reviewed `fe27ac6a44c4560b4201f0b7c75de34237476f04` and the owned-pool report. The report correctly keeps this draft: W4's frozen profile window is fallback-only, and the observed 34 MiB/executor reserved-memory growth is not evidence of stable long-lived lifecycle behavior. These are already disclosed limitations, not newly discovered regressions.

For the next separately budgeted qualification, I suggest three concrete additions to #6:

1. **Prove dispatch coverage in the CPU plan.** Replay the scheduler/shape metadata into the bucket selector before reserving a GPU. Require the selected profile windows to contain the intended bucket-4/bucket-8 and fallback transitions. Eight *physical* rows currently support only four live rows; a W4 batch with eight live requests cannot exercise that bucket. For live-gate workloads, feasibility must additionally confirm actual coverage. Do not relabel the existing fallback-only window as proof.
2. **Keep the mechanism A/B, add a deployment baseline.** Padded eager versus graph replay isolates launch amortization. A separately controlled comparison against the accepted compact runtime is needed for an adoption claim, with setup/capture amortization, padding fraction, bucket hits/fallback reasons, and arrival-to-token latency. The roughly 2x W1 ratio currently measures only the first comparison.
3. **Define the supported executor lifetime.** The bounded 14-engine envelope is legitimate evidence, but process-terminal zero memory does not establish steady-state create/close reuse. Prefer one reusable executor per resident model/cache where possible; if replacement/churn is supported, freeze a lifecycle memory plateau criterion and investigate reserved versus live allocation separately. Do not insert shared allocator-cache drops into serving measurements to mask retention.

Our CDB MVP does not yet have a qualified graph path, so this implementation is a reference to learn from, not a result we have surpassed. Preserve the current complete-but-unqualified evidence and use a new contract for any additional run.

hsliuustc0106:
## Owner review — keep draft; agree with the self-assessment, three additions

Keeping this unqualified is right: the W4 replay-profile coverage is genuinely missing and W1 correlations cannot substitute. Three things I'd want addressed before requalification:

1. **The 34 MiB net reservation growth per fresh executor needs a root cause, not a 14-engine bound.** The per-bucket owned `MemPool`s are the natural suspect. Sharing one pool across buckets (and across engines) via `graph_pool_handle` is the standard mitigation and would also shrink the teardown choreography. If growth persists with a shared pool, that is an allocator interaction worth understanding before this path scales.
2. **The bucket ladder problem from #14 is what produced the missing W4 coverage** — 4-live-row buckets exclude the mixed compositions that dominate serving workloads. Fix the ladder first, then re-run W4; re-running W4 against the current buckets would just re-demonstrate the fallback.
3. **Magic constants**: `payload = 96 * model.config.hidden_size + 1980` and friends are tied to Ouro's shape with no visible derivation. Derive from config or document the formula — they will silently break on any architecture change.

Structural suggestion: the 832-line executor mixes mechanism (capture/replay/pool ownership) with instrumentation (phase events, dispatch records, budget accounting). Splitting the observability surface from the execution surface would materially shrink the review surface of the rework.

Also worth stating in the PR body / roadmap: with synchronous gate readback retained, graphs shave launch overhead only. The 2× W1 result is real but is the ceiling for this mechanism until either the bucket ladder covers mixed batches or a lookahead gate exists — and per #19, none is currently available.

hsliuustc0106 (commented):
revmoe jsons and logs, move capture related benchmarks and vlitations out of the src to a specific subfilder under benchmarks

hsliuustc0106:
## Owner review

**Head 8f3590f, base codex-m3-persistent (clean stack, merge-base = base tip). 48 files, +6980/−268.**
Coverage: full critical pass over the runtime core (recurrent_graph, capture_resources, decode_buffers, graph_diagnostics, model_runner, kv_cache_manager decode transactions, llm_engine, ouro) + testing/maintainability/security/performance specialists + red team + adversarial agent, key findings re-verified by direct code reads; CPU suite run at head in an isolated worktree. Codex cross-model passes unavailable locally (broken CLI install), so independent coverage came from two adversarial passes.

```
Scope Check: CLEAN
Intent: opt-in FP32 graph-capture executor for one recurrent traversal (bucket ladder,
  shared pool + capture-stream lease), runtime under vllm_lt/worker/, ~50% test/benchmark
  footprint reduction, benchmarks/capture v2 contract, docs.
Delivered: exactly that. All claimed components present; benchmarks/ is excluded from the
  wheel as claimed (pyproject includes vllm_lt* only).
Validation re-run locally (macOS): 784 passed / 60 skipped / 1 failed at head — the failure
  is test_benchmark_ab_schema.py::test_active_numa_policy_is_distinct_from_cpuset_allowance
  (reads /proc/self/status, pre-existing on macOS, untouched by this diff). Consistent with
  the PR's 796-pass Linux claim.
```

**Review: 14 issues (1 P1, 3 P2, 10 P3)**

### P1 — silent wrong results

**[P1] (9/10, confirmed independently by three passes) `vllm_lt/worker/recurrent_graph.py:370-385` — setup verification never compares warmup vs replay.** After capture, the verification replay is checked only for `isfinite` and inactive-row zeros/signbits. `warmup_sha256`, `replay_sha256`, and `max_abs_diff` are computed into `bucket["verification"]` (lines 382-384) but a repo-wide grep shows **zero consumers** — the offline audit (`benchmarks/capture/validation.py:642-647`) also checks only `finite` and `inactive_positive_zero`. A replay producing different-but-finite live-row output (wrong Triton specialization captured, stale buffer addresses, pool aliasing) passes setup and every audit, silently corrupting all subsequent decode. Engine-level users of `_enable_recurrent_graph` get no signal; only the separate A/B numerical gate would catch it, and the current head isn't device-qualified anyway (evidence tied to af3327d).
Fix: assert `max_abs_diff` within a float32 tolerance at setup (or hash equality where determinism is guaranteed), and consume the check in the offline audit. The docs (m3-cuda-graphs.md:127-129) accurately describe current behavior, so this is a code gap, not a docs overclaim.

### P2 — hot-path host overhead (all verified by direct read)

The PR's purpose is eliminating per-step host overhead; these three sit directly on the replay path and can rival the graph-launch cost they surround. Historical speedups (1.24-2.13x) were measured *with* these present, so they're headroom, not regressions.

**[P2] (8/10) `vllm_lt/core/kv_cache_manager.py:577,584,616` — the immutable host batch is fully re-validated three times per traversal.** `_validate_decode_host` (O(rows x layers) `_require_prefix` scan, ~96 calls/pass at 24 layers x 4 rows, plus address-set rebuild) runs at begin, bind, and commit; `_require_live_batch` likewise repeats at bind/commit/release. The host batch is frozen tuples in a single-threaded engine — nothing can change between passes. Fix: validate once at begin, stamp a generation on the ticket, and have bind/commit check ticket state plus cheap identity only.

**[P2] (8/10) `vllm_lt/core/kv_cache_manager.py:479,505-523` — six synchronous H2D copies from unpinned staging per decode step.** `_prepare_into` fills six CPU tensors element-by-element (including an inner loop over block-table columns) then issues six `copy_(staging, non_blocking=False)` from `pin_memory=False` memory — each blocks the host until the DMA completes. Fix: one packed pinned staging buffer, a single async copy, stream-ordered before replay.

**[P2] (8/10) `vllm_lt/worker/recurrent_graph.py:420-432` (called at :492) — `_check_bucket` rebuilds ~12 full tensor signatures on every replay.** `id/data_ptr/shape/stride/dtype/device` tuples over the pool x2 plus every bucket tensor, plus a view-identity loop — all invariants fixed at setup for executor-private storage. Fix: keep the cheap `is` identity checks on the hot path; verify full signatures once at setup (or on an epoch bump).

### P3

**[P3] (9/10) `vllm_lt/worker/model_runner.py:89-91` + `vllm_lt/engine/llm_engine.py:39-42` — non-budget setup failure bricks the engine with no reachable recovery.** `_decode_executor` is installed *before* `setup()`; only `CaptureBudgetExceeded` gets the decline path. Any other failure (OOM during capture, Triton compile error) leaves the failed executor permanently installed; `LLMEngine` exposes `_enable_recurrent_graph` but no close, so the retain-for-recovery design (docstring at model_runner.py:88) has no engine-level recovery, and a dropped engine never returns its leased setup stream to `capture_resources._IDLE_STREAMS` (:21-38) nor releases the pool except by GC. Fix: expose close on the engine (or route non-budget failures through the same cleanup attempt).

**[P3] (9/10, deliberate fail-closed) `vllm_lt/engine/llm_engine.py:87-109` — a `_update` failure after successful execute settles and permanently fails a healthy executor.** Post-gate bookkeeping errors (finalize_token, enqueue, output construction) kill the executor and abort all in-batch requests where the compact engine survives the identical error. The comment explains why (pages return to the pool while device work may be in flight) — defensible, but it escalates any host-side bug to engine death; worth a line in the docs.

**[P3] (8/10, latent) `vllm_lt/worker/recurrent_graph.py:476-478` + `vllm_lt/worker/model_runner.py:311` — empty dispatch breaks the completion pairing.** `recurrent()` early-returns "ready" for count==0 while `_execute_batch` calls `complete_after_gate()` unconditionally after RECURRENT → "no recurrent traversal to complete" → executor failed. Unreachable via the engine today only because `torch.stack([])` at model_runner.py:301 raises first. Fix: skip completion for empty dispatches.

**[P3] (9/10) `benchmarks/capture/report.py:714` — "separate setup accounting" is a self-comparison.** `ExecutionAdapter.finish(..., setup_ns=...)` (runtime.py:189-206) feeds the same value into both `result["setup_ns"]` and `capture["setup_ns"]`, so the audit's `equal()` can never fail; the genuinely independent measurement (`setup_record.finished_ns - started_ns`) is never reconciled against it.

**[P3] (8/10) The three `_check_bucket` use-after-free guards are unreachable in the test suite** (`recurrent_graph.py:420-432` — pool identity, storage signature, view metadata): every test keeps the same tensors. Given the P1, these are precisely the guards worth cheap CPU-fake tests for (mutate tensor identity post-setup, assert the specific RuntimeError and no replay). Same for `record_cleanup_failure` routing (recurrent_graph.py:586, model_runner.py:239) and `close()`-while-in-flight. The PR's 50.9% test reduction was a stated goal — these aren't redundant coverage, they're the only tests for the guards that exist.

**[P3] (7/10) Sync-coupling literal clusters** — values that must be edited in lockstep across files: warmup counts 3/1/4 (`recurrent_graph.py:352` / `schema.py:188-190` / `validation.py:528-547`); watchdog `600 * 10**9` x4 in `report.py:504-576` vs the contract's `case_timeout_s` (m2.py already reads it properly); projection id `'loop_gate_logits_full_kv_v1'` x3 (`validation.py:18`, `report.py:142`, `runner.py:381`); pair-prefix `'M2-'→'M3-capture-'` rewrite (`schema.py:142` + `report.py:45`); odd-row interleave x5 sites; scratch 4-pages/2-tokens x5 sites. Each is a named constant away from being unbreakable.

**[P3] (8/10) `benchmarks/capture/runtime.py:32-69` + `validation.py:323,386` — the profiler reaches into `_decode_executor` and `cache._prepare_host_batch` privates and re-implements the production routing decision**; divergence raises only after `original()` already mutated executor/KV state, converting profiling runs into hard failures. A small public read-only surface on the executor would decouple it.

**[P3] (6/10, expected for research tooling) `benchmarks/capture/runner.py:225-231` — plan JSON is a code-executing trusted input** (plan-supplied `interpreter` and roots become worker subprocess cwd/PYTHONPATH; the digest is self-consistency only). Document "never accept plans from third parties" and optionally pin workers to `sys.executable`. Related: `run_worker` catches `BaseException` including KeyboardInterrupt and returns a manifest normally (runner.py:131-135) — acceptable since the controller watchdog owns kills, but Ctrl-C on a local run won't propagate.

**[P3] (9/10) `benchmarks/capture/validation.py:181,199` — typos in frozen messages**: "from15 cases/31 streams", "projected15-case subset" (missing spaces).

**[P3] (standing draft gate) No real-GPU tests at head** — device evidence (49 GPU tests, 17 lifecycle cycles, 2.04-2.13x/1.24-1.30x ratios) is tied to af3327d, seven commits behind head; PR body says so honestly. Plus the standing 512 MiB owned-pool memory-gate run.

### Not flagged (checked, deliberately)

`_restore_scratch` bypassing `free()` (recurrent_graph.py:281-286) — documented refactor tripwire with paired tests; cooperative `setup_timeout_s` and `capture_error_mode="global"` — inherent to the design; `bundle.update(bundle["buckets"][rows])` aliasing; `DecodeBucketLayout.row_counts` rebuild per access.

### Verdict

**Approve-with-comments for when the draft gates pass — fix the P1 first; it's a one-assertion change that closes the PR's one silent-wrong-results hole.** The P2 trio is the obvious next-lever work for M3's performance goal, not a merge blocker. The core engineering itself is careful: close/lease lifecycle, fail-closed settlement, and audit plumbing are all coherently built.

hsliuustc0106:
## Owner review — addendum: file redundancy

Follow-up to the review above at the owner's request: the PR carries substantial file/record redundancy beyond the findings already listed. Concrete inventory, all verified at head 8f3590f.

### Line budget — what this PR actually is

| Category | Added lines | Share |
|---|---|---|
| Production runtime (`vllm_lt/worker` + core/engine/model) | 1,347 | 19% |
| Benchmark/validation tooling (`benchmarks/capture` 2,811 + `vllm_lt/{benchmarks,validation}` 418) | 3,229 | 46% |
| Tests | 2,009 | 29% |
| Docs | 395 | 6% |

Of the 2,009 test lines, ~1,036 (`test_benchmark_capture_{report,runner,runtime,schema}.py` + `test_validation_m3_capture.py`) test the benchmark harness itself, not the engine. **~75% of the PR is harness, harness-tests, and attempt records around a ~1,300-line runtime change.** The PR's own halving of tooling/tests was real and welcome — but the shape is still upside-down relative to the milestone.

### Redundant dated evidence records

`docs/benchmarks/` now holds 11 files, 7 of them M3. This PR adds **four** dated records for one milestone:

- `m3-capture-stopped-20260911.md` (29 L) — attempt 1, failed memory gate, incomplete
- `m3-capture-owned-20260911.md` (56 L) — attempt 2, inconclusive (missing replay evidence)
- `m3-capture-review-20260912.md` (57 L) — attempt 3, gates passed, profiling incomplete
- `m3-capture-review-evidence-20260912/README.md` (59 L) — thin index into the previous file

plus two attempt contracts (`m3-capture-contract-20260911.md`, `m3-capture-owned-contract-20260911.md`) that exist only as pinned blobs at 27f9e80, referenced from the records. And the PR body restates the review doc's headline numbers (2.04–2.13×/1.24–1.30×, 49 GPU tests, 17 cycles) — same facts in three places.

Each new device attempt currently leaves another permanently-dated file describing a failed or superseded run. Recommendation: one `m3-capture` record per milestone state (current attempt in full; earlier attempts collapsed into an attempt-log table: date, source SHA, outcome, one-line cause, checksum pointer), detail archived out-of-tree the way the serving evidence was. PR body should link, not restate.

### Duplicated benchmark logic (verified)

- `benchmarks/capture/report.py:299-486` `_audit_workers` re-implements the worker-manifest audit that already exists in `vllm_lt/benchmarks/ab_report.py` — same launch-order lifetime chain (compare report.py:370-371 with ab_report.py:309-314), same `equal(len(launches), 8, "eight actual launches")`, same adjacent-worker overlap and A1-gate checks. The invariants now exist twice and must be fixed twice.
- `benchmarks/capture/validation.py:20` hardcodes the ladder as `BUCKETS = {"row_counts": [4, 8], "table_width": 32, "max_live_rows": 4, "mapping": "odd"}` and re-implements odd-row arithmetic at :840 — while `DecodeBucketLayout` (the production owner of exactly this) is imported in the same file (:377, :505).
- `benchmarks/capture/runtime.py:46-52` re-implements the executor's dispatch-routing decision from `recurrent_graph.py:456-463`.
- Graph limits exist in three encodings that must agree: `GraphLimits` defaults (`graph_diagnostics.py:31-32`), the frozen fixture (`fixtures/ouro-m3-capture-contract.json`: 1048576/16384), and bare literals in `model_runner.py:52`.

This also deepens the existing split between the two "benchmarks" package roots (`benchmarks/` importing from `vllm_lt/benchmarks/` and `vllm_lt/validation/`). Consolidation direction: shared audit helpers live once in `vllm_lt/benchmarks/ab_report.py`; the auditor derives its expectations from `DecodeBucketLayout`; routing goes behind a small public read-only accessor on the executor.

### Suggested trim (no information loss)

1. Collapse the four dated records → one milestone record + attempt-log table; archive attempt detail out-of-tree with checksums.
2. Derive `BUCKETS`/odd-row expectations in `validation.py` from `DecodeBucketLayout`; extract the shared worker-audit invariants into one helper used by both reports.
3. Single source for graph-limit numbers (fixture generated from, or validated against, `GraphLimits`).
4. PR body: replace restated numbers with links to the record.

None of this blocks the draft gates; it's cleanup that gets cheaper the earlier it's done, and it compounds across the remaining M3/M4/M5 PRs if the per-attempt-file pattern stays.

Issue #7, opened 2026-09-10
M4: improve profile-selected attention and KV bottlenecks


hsliuustc0106:
## Precision requirements — 2026-09-12

For new work, **BF16 is the primary GPU inference, kernel-validation, quality, and performance target**. Use BF16 weights, ordinary activations, and KV storage; specify accumulation independently and retain FP32 in numerically sensitive reductions/intermediates. Full-model FP32 is an optional diagnostic/reference configuration, not the default prerequisite. See the [precision policy](https://github.com/hsliuustc0106/vllm-lt/blob/c9e646ace2d8198a52e522ea4d71129f1b143c58/docs/precision-policy.md).

BF16 profiling, matched optimization A/B, and quality data collection may proceed alongside the original Q1 reference diagnosis after each run's functional, finiteness, and ownership prerequisites pass. Report observations, reference fidelity, and qualification separately. Preserve exact cache/data-movement/RNG invariants and metadata-only A/B behavior; differing arithmetic paths use independently justified, predeclared BF16 numerical and decision/quality criteria. Universal bitwise or top-1 equality is not a blanket mixed-precision requirement, and near-tie exemptions cannot be invented after a run.

For dtype choices, follow explicit original Ouro paper guidance first, then the pinned release configuration and inference code. The [source audit](https://github.com/hsliuustc0106/vllm-lt/blob/c9e646ace2d8198a52e522ea4d71129f1b143c58/docs/paper-notes.md#original-ouro-paper-and-precision-guidance) distinguishes published FP32 operations from backend/project choices; the paper has no per-kernel inference recipe.

This amendment supersedes conflicting dtype/eligibility requirements below for successor work. All dated contracts, raw results, original failures, and historical acceptance decisions remain unchanged. New criteria need a separately frozen contract; no benchmark or qualification is claimed by this edit.

New attention candidates use BF16 inputs/outputs and KV storage, retaining justified FP32 accumulation/softmax statistics. Declare any accumulation change as an isolated variable. Validate kernels on identical inputs with operation/shape-specific bounds and model/decision checks; keep exact KV ownership/copy invariants. BF16 Tensor Core work is not established merely by BF16 storage.

[...]

PR #16, opened 2026-09-11
Evaluate a 64-token Ouro attention tile against M4 guardrails

hsliuustc0106:
Evaluates a 64-token attention tile against the existing 32-token tile on the compact runtime. Stacked on #14.

The candidate failed the frozen M4 performance milestone and remains draft/unqualified. W1 changed -0.1246% and +2.1303% in the two pairs, missing the required improvement in both. The default remains tile32.


hsliuustc0106:
## Owner review — keep rejected; exemplary negative result, two process notes

This is the right kind of failure record: frozen guardrails, both pairs missing the ≥1% requirement with overlapping observed ranges, full evidence published, default unchanged. Two process suggestions:

1. The issue #7-B contract promised "a finite, declared search over tile sizes and warp counts"; this PR searched one point (BLOCK_T=64 at fixed 4 warps). Either run the small declared grid — {32, 64} × {2, 4, 8} warps is six bounded candidates — or record why a single point was chosen, so the contract and the execution match.
2. With two pairs per candidate and observed ranges overlapping by ~2%, a ±1% threshold is underpowered: the −0.1246%/+2.1303% result is consistent with both "no effect" and "small effect" at this sample size. A pre-declared sample size with a power note would prevent spending reserved GPU time on inconclusive tiles. We hit the same discipline question in vllm-loop-serving M2's held-out gate — freezing the candidate and sample before the run was what made a failure interpretable rather than arguable.

Direction suggestion: if the W-profiles show prefill attention as a meaningful share of W4, option A (query-tiled prefill kernel with context-tile reuse) is likely a better next candidate than more decode-tile exploration — decode batches are small (≤8 rows × 16 heads = 128 programs) and the kernel is closer to launch-bound than tile-bound.

PR #17, opened 2026-09-11
feat: add cached Ouro reference adapter

hsliuustc0106:
Adds a cached Ouro reference adapter for FP32/BF16 validation, sharing model weights while retaining independent depth/layer KV histories. Reuses official initialization and accounts for BF16 weights in memory preflight. Stacked on #14.

Prior validation: 698 CPU tests passed; 49 GPU cases skipped, followed by 29 adapter/runner checks. Historical FP32 generation equivalence passed, but its timing controls were invalid; timing remains inconclusive.


hsliuustc0106:
### [P2] Publish timing invalidation in a machine-readable decision layer

Reviewed head `a4430f0031b6c9b46ae44cdff4b1f43352f8dc35`. The PR/roadmap correctly classify timing as inconclusive because the CPU preparation probe overlapped N2 on the same binding. Generation equivalence is a separate, still useful result.

However, `vllm_lt/benchmarks/q2_external_report.py::_finish` still emits `decision="passed", passed=True` for a complete report without its recorded errors/failures, and the PR explicitly says the original automated report passes. A downstream tool consuming only that JSON can therefore promote a result the later chronology review has invalidated for timing.

Please preserve the frozen raw report, but add a **versioned final qualification record** that references its hash plus the chronology evidence and exposes separate fields such as `generation_equivalence=passed`, `timing_controls=invalid`, `performance_qualification=inconclusive`, and `supersedes=<original-report-hash>`. The roadmap/release's canonical result should point to this record. A CPU test should demonstrate that valid token equivalence plus known control contamination cannot yield overall timing qualification.

This needs no additional GPU run and does not require rewriting historical evidence. In our implementation we likewise keep numerical/discrete/milestone decisions separate; a passed subcheck must not stand in for full acceptance. Any future native-versus-official timing round also needs CPU/NUMA workload isolation across preparation and measurement jobs, not only an exclusive GPU reservation. Related issue #8.

hsliuustc0106:
## Owner review — keep draft; generation comparability is the keeper result

The identical 64-token depth-4 histories against the pinned official adapter (shared immutable weights, 96 distinct official depth/layer cache slots, 0/0 final task-owned memory) is the strongest external correctness evidence in either codebase — stronger than logit tolerances for the question that matters: does the paged/last-exited implementation compute what the official engine computes.

On the timing contamination: catching it via post-hoc chronology review is good, but the general fix is a fail-fast pre-run gate — pin and verify the CPU affinity of measurement threads before the first timed region and abort if the isolation assertion fails, rather than discovering the overlap in review. vllm-loop-serving's M0 harness records affinity as nullable and has the same weakness; both repos should treat a run-time isolation assertion as an acceptance gate, not a review finding.

For the eventual re-run: since generation comparability is already established, consider recording the official-adapter timing under the same isolation gate in the same session as N2 — one reservation, paired environment, and the comparability check comes free with the timing.

hsliuustc0106:
Profiled the BF16 native/HF cached-adapter comparison at head `c473654c548fee8db7bdeb40e562f5fd02efe19b`. Native issued **9.45% fewer decode kernels**, but accumulated **6.40% more GPU kernel time** in this capture. Reduced dispatch overhead is a plausible contributor to the earlier native latency advantage; these traces do not establish causality or a new speedup result.

One capture per backend followed one finite feasibility/warmup request per backend: **four requests total, no retries**. Each request used 128 input tokens, 64 greedy outputs and four loops. PyTorch CPU/CUDA profiling covered prefill plus the first 16 decode outputs; the remaining 47 outputs completed outside the capture.

The table covers the **16 captured decode outputs**:

| Metric | vllm-lt native | HF cached adapter |
|---|---:|---:|
| Kernel launches | 88,816 | 98,080 |
| Kernel launches per output | 5,551 | 6,130 |
| Summed GPU kernel duration | 219.62 ms | 206.41 ms |
| Summed kernel duration per output | 13.73 ms | 12.90 ms |
| Pointwise, cast and reduction kernel duration | 98.31 ms (44.76%) | 107.27 ms (51.97%) |
| CUDA stream synchronization calls | 416 | 48 |
| Recorded copy events | 560 | 112 |
| Summed device copy duration | 0.612 ms | 0.138 ms |

Suggested optimization follow-ups:

1. **Fuse RMSNorm and adjacent pointwise operations, preserving FP32 intermediates.** Both backends record 6,208 mean reductions, or 388 per decode output, alongside separate casts, squares, additions, reciprocal square roots and multiplies. The broader pointwise/cast/reduction group accounts for 45–52% of summed kernel time; that percentage is not RMSNorm alone. The [native RMSNorm implementation](https://github.com/hsliuustc0106/vllm-lt/blob/c473654c548fee8db7bdeb40e562f5fd02efe19b/vllm_lt/models/ouro.py#L33) exposes a concrete fusion candidate.
2. **Tune native paged attention.** Its 1,536 launches take 29.09 ms, or 13.24% of native kernel time. The [current kernel](https://github.com/hsliuustc0106/vllm-lt/blob/c473654c548fee8db7bdeb40e562f5fd02efe19b/vllm_lt/kernels/triton_attention.py#L15) uses a 32-position tile and FP32 arithmetic. HF's batched matmul/softmax events alone are not a complete attention baseline because other attention operations contribute additional work.
3. **Reduce native KV-write and metadata submissions.** The [compact path](https://github.com/hsliuustc0106/vllm-lt/blob/c473654c548fee8db7bdeb40e562f5fd02efe19b/vllm_lt/core/kv_cache_manager.py#L542) performs separate indexed K/V writes: 3,072 kernels totaling 10.50 ms. Native also records 26 stream synchronizations per output versus HF's 3. Device copy time is small, making submission/synchronization overhead a candidate for investigation.

HF records 6,176 concatenation kernels versus native's 3,136 (11.36 ms versus 5.31 ms). The [HF cache](https://github.com/hsliuustc0106/vllm-lt/blob/c473654c548fee8db7bdeb40e562f5fd02efe19b/vllm_lt/validation/reference_code/modeling_ouro.py#L177) appends K/V with concatenation, but both backends also concatenate rotary components, so total `cat` time must not be labeled exclusively as KV-cache time. CUDA Graph replay is another dispatch-overhead candidate; its benefit was not measured here.

Controls and validation: shared immutable BF16 Ouro-1.4B weights at model revision `574fa66cb8bf5abdc979642d01cf2b79b16bfab1`; Transformers 4.55.0 pinned cached adapter, rather than `AutoModel.generate()`; one thread; TF32 and reduced-precision matmul reductions disabled; the same 3 GiB native pool resident for both; CUDA Graphs and persistent decode disabled. All four 64-token histories matched the earlier BF16 output exactly. HF retained 96 distinct cache slots at final length 191. The worker exited, task-owned allocated/reserved CUDA memory returned to zero, and its reservation was released.

**Measurement limits:** both configurations used the same free L20X GPU4, CPU68 and NUMA1 because the earlier GPU2 was occupied. These captures therefore use different placement from the original benchmark. Profiling adds substantial CPU overhead, kernel totals are not end-to-end latency, and there is only one capture per backend. HF prefill also contains a 282.02 ms CPU self event named `VmappedVmappedVmappedVmappedModIndex`; this warrants investigation but does not establish a prefill regression. Keep these diagnostics separate from the [uninstrumented BF16 comparison](https://github.com/hsliuustc0106/vllm-lt/blob/c473654c548fee8db7bdeb40e562f5fd02efe19b/docs/benchmarks/q2-external-bf16-20260912.md).

Raw traces, summaries, protocol and external harness are retained locally; they are not attached to this comment. Raw trace SHA256: native `d492bde5a83ea808ad1721e700e41cdf80d7000b175c540515aea1aa0ca8ce95`; HF `c03abeee99da2c426bd609b9cbe7f07f917beebde69ffa50f5b1412ad8c57cb4`. Repository source was unchanged for profiling. These are optimization candidates, not merge-blocking findings.

PR #18, opened 2026-09-11
Add a bounded FP32 Ouro quality baseline (Q2)

hsliuustc0106:
Adds deterministic FP32 quality selection, a bounded actual-generation controller and an offline scorer. Stacked on #14.

Prior validation: 128 CPU tests passed. The real-weight screen scored 4/64, with 58 unparseable answers and 57 length-limit terminations. The score is unchanged; this draft does not qualify task quality or performance, and longer-history diagnosis remains open.


hsliuustc0106:
### Quality review: first distinguish the evaluation protocol from engine fidelity

Reviewed head `43fdeee6620af58cd274e2ca3876fbffd3808e15`, the frozen template/parser, and the result summary. Keeping the 64-example denominator, 58 unparseable answers and 57 length terminations is correct. The observed 4/64 is the result of this **raw-template, 256-output-token, strict-parser protocol**; it does not by itself localize a native-engine defect or establish the checkpoint's general GSM8K accuracy.

For a new bounded diagnostic under #8:

- Run native and pinned official cached inference on identical tokenized inputs, output budget, dtype and fixed-depth policy for a small predeclared set. PR #17's W1 equivalence is useful, but does not cover these longer reasoning histories. Capture first differing tokens/logits if they disagree.
- Audit prompt formatting/BOS/EOS against the pinned checkpoint's intended interface. `q2_quality_data.py` currently uses a raw `Question: ... Answer:` template with `add_special_tokens=False`; this is a declared choice, not automatically a bug. Compare a checkpoint-appropriate template only as a separately labeled experiment.
- Inspect the already-preserved text to classify parse failures and token-limit terminations. A separately planned longer-output diagnostic can distinguish truncation from answer-format compliance; it must not retroactively rescore or replace the primary 4/64 result. Keep engine fidelity, format compliance and mathematical correctness as distinct outcomes.

Our serving MVP's smoke/teacher-forced checks also do not establish task quality, and its BF16 exact-token issue remains open. Neither engine should use agreement on forced histories as a substitute for this free-generation check. Prioritize this diagnosis before learned lookahead: a gate-quality comparison built on an unexplained low baseline would be difficult to interpret.

hsliuustc0106:
## Owner review — keep draft, but the headline is probably a prompt-format artifact, not model quality

Engine correctness is not in question — #17's identical histories against the official adapter covers that. The result signature here (4/64 correct, **58/64 unparseable, 57/64 token-limit terminations**) is the classic shape of a prompt/template mismatch rather than a broken engine or a globally weak model:

- If the checkpoint expects its chat template (or the model card's documented format) rather than a plain question-only completion prompt, greedy decoding produces well-formed tokens that never settle into the `#### <number>` answer format — which is exactly 58-unparseable + 57-truncated.
- Reference point: vllm-loop-serving's real-checkpoint smoke (same pinned revision, plain prompt, greedy, fixed depth four) produced a correct direct answer on an arithmetic prompt — basic competence is present.

Cheap decisive experiments before any quality conclusion is drawn (both fit one reservation with the existing #17 adapter):

1. Run the same 64 prompts through the **official adapter with the identical raw prompt**. If it also scores ~6% with ~58 unparseable, the prompt is the variable, not the engine — and the baseline becomes "quality under this prompt", a different (still publishable, much less alarming) claim.
2. Ablate prompt format on a small fixed subset (plain vs chat template vs model-card format), frozen before inspection like the rest of the protocol.

Also reconsider the 256-token cap for this reasoning format — GSM8K-length chains are borderline at 256, and truncation conflates "cannot finish reasoning" with "will not parse". The frozen-denominator discipline throughout is excellent; this comment is only about which variable the 6.25% actually measures.

Issue #9, opened 2026-09-10
M5: assess and qualify asynchronous lookahead routing


hsliuustc0106:
## Precision requirements — 2026-09-12

For new work, **BF16 is the primary GPU inference, kernel-validation, quality, and performance target**. Use BF16 weights, ordinary activations, and KV storage; specify accumulation independently and retain FP32 in numerically sensitive reductions/intermediates. Full-model FP32 is an optional diagnostic/reference configuration, not the default prerequisite. See the [precision policy](https://github.com/hsliuustc0106/vllm-lt/blob/c9e646ace2d8198a52e522ea4d71129f1b143c58/docs/precision-policy.md).

BF16 profiling, matched optimization A/B, and quality data collection may proceed alongside the original Q1 reference diagnosis after each run's functional, finiteness, and ownership prerequisites pass. Report observations, reference fidelity, and qualification separately. Preserve exact cache/data-movement/RNG invariants and metadata-only A/B behavior; differing arithmetic paths use independently justified, predeclared BF16 numerical and decision/quality criteria. Universal bitwise or top-1 equality is not a blanket mixed-precision requirement, and near-tie exemptions cannot be invented after a run.

For dtype choices, follow explicit original Ouro paper guidance first, then the pinned release configuration and inference code. The [source audit](https://github.com/hsliuustc0106/vllm-lt/blob/c9e646ace2d8198a52e522ea4d71129f1b143c58/docs/paper-notes.md#original-ouro-paper-and-precision-guidance) distinguishes published FP32 operations from backend/project choices; the paper has no per-kernel inference recipe.

This amendment supersedes conflicting dtype/eligibility requirements below for successor work. All dated contracts, raw results, original failures, and historical acceptance decisions remain unchanged. New criteria need a separately frozen contract; no benchmark or qualification is claimed by this edit.

When gate/runtime experiments become applicable, target BF16 with explicitly justified accumulation precision. Gate availability, calibration, independent history semantics, same-gate runtime behavior, and lifecycle prerequisites remain in force. This revision does not authorize training or an asynchronous implementation.

[...]

PR #19, opened 2026-09-11
docs: move M5 assessment out of the source tree

hsliuustc0106:
Records the M5 assessment outcome in the external archive and removes intermediate planning documents from the source tree. The inspected sources did not contain a compatible published Ouro lookahead gate; asynchronous implementation remains deferred under #9.

Assessment only: no runtime implementation, training or GPU comparison. The availability finding is scoped to the original inspected revisions; learned-gate quality and asynchronous runtime acceptance remain open.


hsliuustc0106:
## Owner review — approve

This is the right outcome shape for a feasibility gate: a documented no-go (no compatible published lookahead gate in the inspected scope) with a bounded proposal parked behind explicit authorization, and no speculative runtime built.

Cross-repo note: vllm-loop-serving's execution contract reached the same conclusion from the other direction — the paper's replayed-exit/shared-KV speed setup is a different configuration from live-gate last-exited semantics, and the stock gate cannot be reinterpreted as a lookahead without a justified mapping, which is exactly this assessment's framing. If the training/calibration proposal is ever pursued, two requirements from our side:

1. Calibrate the trained head against the stock cumulative-hazard gate as the baseline arm, so any overlap benefit attributes to the lookahead rather than to the gate change.
2. Pin the target semantics in the contract before any data work — conditional hazard vs next-step cumulative exit probability, which normalized hidden state enters the head, and how residual probability is initialized/updated (the checklist already in issue #9).

Issue #22, opened 2026-09-12
feat: add OpenAI-compatible serving and reuse vllm bench serve


hsliuustc0106:
## Problem and goal

vllm-lt has an offline M1 engine benchmark, but no HTTP server. We cannot yet measure end-to-end serving throughput and latency under controlled request rates and concurrency using `vllm bench serve`.

Add a minimal OpenAI-compatible serving frontend for Ouro and reuse the upstream benchmark client unchanged. Keep the M1 harness for loop-depth, scheduler, KV, and profiler analysis; HTTP measurements include a different timing boundary.

Related: #2 (serving frontend roadmap), #3 (offline benchmark). This frontend can wrap the existing synchronous engine; it does not require the learned lookahead gate or asynchronous GPU routing in #9.

## Proposed scope

- Serve one resident model on one device. Use BF16 as the primary GPU target under the [current precision policy](https://github.com/hsliuustc0106/vllm-lt/blob/c9e646ace2d8198a52e522ea4d71129f1b143c58/docs/precision-policy.md), with accumulation precision specified independently.
- Implement `/v1/completions`, `/v1/models`, and `/health`. Support text prompts, ordinary and SSE streaming responses, incremental text deltas, finish reasons, accurate prompt/completion token usage, `stream_options.include_usage`, and the terminal `[DONE]` event. Preserve tokenizer behavior across partial text and special tokens.
- Map supported generation parameters to `SamplingParams`, including `max_tokens`, temperature, top-p/top-k, seed, and `ignore_eos`. Expose Ouro loop bounds and exit threshold as documented extension fields usable through the client's `--extra-body`. Honor neutral/default fields emitted by the pinned client, such as `repetition_penalty=1` and `logprobs=null`; reject unsupported behavior explicitly.
- Give one worker exclusive ownership of `add_request()`, `step()`, and `abort_request()`. Accept new requests between steps so independent HTTP requests share continuous batching. Keep the HTTP event loop responsive while engine work runs; the engine is synchronous and not thread-safe.
- Bound pending requests and slow-client output buffering, define overload responses, and route disconnects/cancellation through the engine owner. Cover EOS, errors, shutdown, and request/KV cleanup without affecting other requests.
- Mark health ready only when model, tokenizer, and engine initialization have completed. Validate the first actual inference after readiness and report preparation separately from process-to-readiness.

[...]

PR #23, opened 2026-09-12
feat: add OpenAI-compatible Ouro serving

hsliuustc0106:
Adds a single-model Ouro completions server with real readiness, streaming, bounded queues, shared engine ownership, cancellation and shutdown handling. Supports the unmodified vLLM benchmark client. Implements #22.

Prior validation: 216 CPU tests passed; 23 optional/GPU cases skipped, plus 49,413 tokenizer comparisons. Earlier reserved-device checks passed four direct/HTTP comparisons and 48 client requests; later serving changes have CPU validation only. No speed claim. Authentication/TLS, chat, stop strings and multiple completions remain unsupported.


hsliuustc0106:
## Review — approve-with-comments (no P1)

Reviewed at head `8142907` (13 files, +1780/−1): full diff read, scheduler/engine/protocol checked against callers outside the diff, CPU suite rerun locally (54/54 passed), plus six independent review passes (testing, maintainability, security, performance, API-contract specialists + red team; 31 raw findings deduplicated below).

**Scope:** clean against #22 — all five acceptance criteria are covered by code, tests, evidence, and docs. The BF16 serving target complies with the 2026-09-12 precision policy.

### P2 — recommend before merge

1. **A decode anomaly takes the whole service down, contradicting the documented failure isolation.** `IncrementalText.decode` raises `ValueError` on non-prefix-stable output (`vllm_lt/serving/protocol.py:105`), but `_tick` wraps only `encode`/`add_request` — the `decoder.decode(...)` call at `vllm_lt/serving/worker.py:136` is outside the try. Reproduced with a non-prefix-stable tokenizer: one bad request hits `_run`'s blanket `except Exception`, `worker.ready` goes False permanently, all in-flight requests get 503 `engine_unavailable`, and `/health` returns 503 — contradicting docs/serving.md's "Invalid requests, slow consumers and cancellation preserve unrelated engine work." Unreachable with the compliant byte-level tokenizer, but the guard exists precisely for the misbehaving case and should fail one request, not the server. Fix: contain decode errors per request (400 + abort), like the encode path, and add the negative-path test.

2. **`EngineWorker.close()` leaks `asyncio.TimeoutError` and breaks the shutdown sequence on the hung-device path.** `vllm_lt/serving/worker.py:106` awaits the driver task under `wait_for`; when a tick outlives the deadline, `close()` raises. aiohttp 3.12's `BaseRunner.cleanup()` runs `on_shutdown` (→ `stop(app)` → `worker.close()`) *before* handler drain and `on_cleanup`, so the escaping exception skips both and `web.run_app` exits with a traceback — exactly the "device call never returns" scenario the docs delegate to an outer supervisor. Fix: catch `TimeoutError` inside `close()` and log "shutdown deadline exceeded; forcing teardown". Related: after `executor.shutdown(cancel_futures=True)`, `_run`'s finally `_unload` raises "cannot schedule new futures" — make that path log cleanup-skipped.

3. **docs/serving.md cites the precision policy at a commit that exists on no mergeable branch.** Line 22 pins `docs/precision-policy.md` at `c9e646a`, which is diverged from `main` and from every branch — the file currently ships only in #12. If #23 merges first, main carries a "current precision policy" reference to a document not in the tree, and the absolute pinned URL won't track the file once #12 lands. Suggest landing #12 first, then switching to a relative link.

### P3 — informational

- **Scheduler guard changes offline refill-mode interleaving** (`vllm_lt/core/scheduler.py:116`, `:140`): a RECURRENT step now lands between consecutive PREFILL batches whenever prompts span more than one `max_num_batched_tokens` batch. Per-request outputs are unchanged (greedy, per-request RNG), but wall-clock step composition differs from the frozen M1 evidence on rerun. One sentence in docs/benchmarks would cover it.
- **O(n²) detokenization on the serialized owner thread** (`vllm_lt/serving/protocol.py:98`): `decode()` re-decodes the full cumulative token list every generated token, bounded by context length but stalling co-batched requests as outputs grow. Byte-level tokens decode independently — decode only the new suffix.
- **Browser-origin exposure beyond the declared scope** (`vllm_lt/serving/server.py:87`): `request.json()` ignores Content-Type and Host/Origin are unchecked — cross-site "simple" POSTs from a web page can drive GPU work or starve slots, and DNS rebinding can read completions. A Host + Content-Type middleware check closes this for a localhost tool.
- **Validation script under-enforces the documented claims** (`scripts/validate_serving.py:343`): the docs claim finite/nonnegative TTFT/ITL and a clean process exit; the script asserts counts only. Assert finiteness/nonnegativity and `server_returncode == 0`, and record first-request latency.
- **Undisclosed API deviations from OpenAI semantics:** missing `model` returns 404 (OpenAI: 400); `error.type` mirrors `error.code`; explicit JSON `null` for `SamplingParams`-backed fields returns 400 while neutral fields accept null; 400 messages leak KV-block internals. Align or disclose in docs/serving.md.
- **Test gaps (coverage-verified):** the documented 400 paths for context overflow (`vllm_lt/engine/llm_engine.py:56`) and loop bounds (`:53`) are never exercised end-to-end; the cancel-before-first-tick branch (`vllm_lt/serving/worker.py:126`) is untested; `load_engine` and `scripts/validate_serving.py` have zero coverage.
- **Maintainability:** default limits and error tuples are tripled across worker.py/server.py/serve.py (drift risk); `decoder.__getstate__()` introspects tokenizers' private payload (`vllm_lt/entrypoints/serve.py:26` — `isinstance(decoder, tokenizers.decoders.ByteLevel)` is the public check); validate_serving asserts on the exact worker log string.
- **Stale PR body:** it still says raw results are in `docs/benchmarks/serving-20260912/`, but `8142907` moved evidence out of the tree to ignored `artifacts/`.

### Verified good

Parameter validation is genuinely default-deny (bool/NaN/unknown-field handling correct); request/KV slots are released through the owner on disconnect/timeout; no subprocess/shell injection; all resource bounds present (body size, request count, output buffer, write/request/shutdown timeouts); SSE framing matches the pinned client's parser (15 ITL samples, no phantom token events); evidence claims in docs/benchmarks/serving-20260912.md match the script's actual assertions where they overlap.

PR #31, opened 2026-09-17
[feat] add NIXL PD disaggregation and KV scheduling features

bjf-frz:
## Purpose

  Add NIXL-based prefill/decode disaggregation, together with KV cache management, scheduling, and Prefill metadata optimizations.

  | Feature | Implementation | Notes |
  |---|---|---|
  | Multiprocess PD serving | Run Prefill and Decode in separate worker processes on configurable GPU groups. | Supports ratios such as 1P3D, 2P2D, and 3P1D. Prefill executes all recurrence depths; Decode supports per-request early exits. |
  | NIXL KV transfer | Transfer paged KV and the final prompt hidden state from Prefill to Decode. | Bounded transfer chunks and in-flight bytes allow transfer to overlap with Prefill progress. |
  | Transfer-safe KV lifecycle | Protect in-flight KV pages with transfer leases and defer reclamation until transfers finish. | Decode requests remain in a receiving state until the required data is ready. |
  | Separate compute and transfer capacity | Release Prefill compute capacity independently of transfer completion. | Bounded receiving and draining slots provide backpressure. |
  | Prefix caching | Reuse immutable, complete prompt blocks across all recurrence depths. | Supports `last_exited` KV, reference counting, eviction, and publication after GPU completion. Recomputes the final prompt token to recover its hidden state. |
  | Prefix-aware PD transfer | Reuse local prefixes on both workers and transfer only the suffix missing on Decode. | P→D transfer only; no reverse cache fetching or global cache-aware routing. |
  | Incremental KV allocation | Grow physical KV allocations as requests advance. | Updates resident GPU block tables. PD admission still conservatively reserves logical Decode capacity. |
  | KV admission watermark | Maintain configurable free-page headroom during admission. | Enabled through `--kv-watermark`. |
  | Priority scheduling | Support request priorities alongside FCFS scheduling. | Smaller values take precedence; priority preemption requires preemption to be enabled. |
  | Lossless preemption | Preserve KV and request state on CPU, then restore them to resume execution. | Uses snapshots instead of recomputation to preserve looped execution history. Incurs CPU memory and transfer costs. |
  | UVA-backed Prefill metadata | Use pinned host buffers and Triton kernels to prepare GPU metadata. | Requires CUDA FA4 with `last_exited` KV. Event-protected reusable buffers; no vLLM dependency. |
  | Serving lifecycle integration | Handle pending KV reception, cancellation, worker failures, and coordinated shutdown. | Receiving-only states no longer trigger scheduler progress errors. |
  | Configuration and entrypoints | Add a PD serving entrypoint, optional NIXL dependency, and runtime flags for the new features. | Features are explicitly configurable. |
  | Regression tests | Cover PD execution, transfer ownership, cleanup, prefix reuse, incremental allocation, preemption, and UVA metadata. | Includes CUDA/NIXL tests when hardware and dependencies are available. |

## Test Plan

  - Evaluate end-to-end performance:
    - Compare four independent replicas with 1P3D, 2P2D, and 3P1D, using four GPUs per configuration.
    - Replay the same 512 ShareGPT prompts at 4 and 8 requests/s, with 128 output tokens per request.
    - Measure initial requests and immediate replay separately; replay does not imply full cache hits.
    - Repeat measurements with exchanged GPU groups.
    - Record throughput, TTFT, TPOT, ITL, E2E latency, request failures, and latency-qualified Goodput across a threshold grid.
    - Independently rerun anomalous 1P3D and four-replica cases without a concurrent benchmark.
    - Retain per-request timestamps, server logs, and original anomalous measurements.

  ## Test Results

  Four-GPU comparison using Ouro-1.4B BF16, FA4 (split=1), asynchronous scheduling, CUDA Graphs, and the new KV/scheduling features. Each phase uses 512 ShareGPT prompts with 128 output tokens per request.

  ### Averaged performance

  Each row averages the initial-run and replay metrics with equal weight. Latency values are averages of phase percentiles, not request-mean latencies or percentiles recomputed across both phases.

  Percentages show changes relative to four replicas at the same arrival rate. Higher throughput is better; lower latency is better.

  | Load | Configuration | Throughput (tok/s) | TTFT P95 (s) | TPOT P95 (ms) | ITL P99 (ms) | E2E P95 (s) |
  |---|---|---:|---:|---:|---:|---:|
  | 4 req/s | 4 replicas | 503.4 | 0.650 | 36.7 | 114.5 | 5.05 |
  | 4 req/s | 1P3D | 501.2 (-0.5%) | 0.743 (+14.3%) | 24.6 (-32.9%) | 34.5 (-69.8%) | 3.67 (-27.3%) |
  | 4 req/s | 2P2D | 498.3 (-1.0%) | 0.563 (-13.3%) | 33.2 (-9.5%) | 53.2 (-53.5%) | 4.68 (-7.3%) |
  | 4 req/s | 3P1D | 497.7 (-1.1%) | 0.579 (-10.9%) | 45.3 (+23.6%) | 65.4 (-42.9%) | 6.14 (+21.6%) |
  | 8 req/s | 4 replicas | 976.5 | 1.202 | 49.5 | 123.5 | 7.01 |
  | 8 req/s | 1P3D | 958.6 (-1.8%) | 1.848 (+53.8%) | 29.9 (-39.6%) | 46.8 (-62.1%) | 5.21 (-25.6%) |
  | 8 req/s | 2P2D | 951.9 (-2.5%) | 0.892 (-25.8%) | 43.6 (-11.8%) | 64.3 (-47.9%) | 6.14 (-12.4%) |
  | 8 req/s | 3P1D | 872.4 (-10.7%) | 4.848 (+303.3%) | 46.5 (-6.0%) | 70.1 (-43.2%) | 9.06 (+29.3%) |

  The 1P3D results at 4 req/s and four-replica results at 8 req/s use independent reruns without a concurrent benchmark. Other results aggregate two repetitions collected with two four-GPU configurations running concurrently on the same host. Original anomalous measurements are retained separately. This is a consolidated comparison, not a uniformly isolated benchmark.

  ### Analysis

  #### 1P3D: faster generation, with a first-token latency tradeoff

  At 4 req/s, 1P3D reduces TPOT by 32.9%, ITL by 69.8%, and E2E latency by 27.3%, while TTFT increases by 14.3%.

  At 8 req/s, TPOT decreases by 39.6%, ITL by 62.1%, and E2E latency by 25.6%. However, TTFT increases by 53.8%.

  This pattern is consistent with dedicating three GPUs to Decode while concentrating Prefill on one GPU. Decode avoids competing with Prefill execution, improving generation speed and reducing interruptions. Meanwhile, the single Prefill worker handles all incoming prompts, potentially increasing the wait for the first token.

  For this workload, 1P3D favors generation responsiveness over first-token latency. It does not improve every metric.

  #### 2P2D: the most balanced configuration

  2P2D improves all reported latency metrics at both arrival rates.

  At 8 req/s, TTFT decreases by 25.8%, TPOT by 11.8%, ITL by 47.9%, and E2E latency by 12.4%, with 2.5% lower measured throughput.

  Two Prefill workers provide more prompt-processing capacity than 1P3D, while two Decode workers retain sufficient generation capacity for the tested load. The observed tradeoff is better TTFT but slower generation than 1P3D.

  Among the tested ratios, 2P2D offers the best balance between first-token responsiveness and generation smoothness. This conclusion applies to the measured input distribution and fixed 128-token outputs.

  #### 3P1D: Decode-side pressure limits the benefit

  At 4 req/s, 3P1D improves TTFT by 10.9%, but TPOT increases by 23.6% and E2E latency by 21.6%.

  At 8 req/s, throughput falls by 10.7%, TTFT increases by 303.3%, and E2E latency increases by 29.3%. The remaining 6.0% TPOT improvement cannot compensate for the longer first-token wait.

  This behavior is consistent with a Decode-side bottleneck. Additional Prefill capacity cannot compensate for requests waiting for the single Decode worker. Admission limits, KV capacity, and Decode execution time may all contribute; their individual impact is not isolated by these measurements.

  Lower ITL alongside worse E2E is possible because requests can wait longer before generation starts while experiencing fewer interruptions after generation begins.

  #### Throughput: limited by the offered load

  The offered output workload is approximately 512 tok/s at 4 req/s and 1,024 tok/s at 8 req/s. Configurations that keep up with arrivals naturally report similar throughput.

  Throughput includes the time needed to drain outstanding requests after the final arrival. Longer drain time lowers measured throughput even when all requests succeed.

  These results therefore compare fixed-load service behavior, not peak throughput. The small throughput differences between four replicas, 1P3D, and 2P2D do not establish their maximum sustainable capacity.

  #### ITL: the most consistent observed PD improvement

  Across the tested ratios, ITL P99 decreases by approximately 43–70% at 4 req/s and 43–62% at 8 req/s.

  This is consistent with separating Decode from Prefill execution. Per-request TPOT averages can hide individual token-generation stalls, whereas ITL directly exposes those interruptions.

  The ITL reduction is meaningful even when throughput changes little. However, it must be evaluated alongside TTFT and E2E, particularly for 3P1D.

  #### Attribution and measurement limitations

  All new features are enabled together. These results compare deployment configurations and do not isolate gains from Prefix Cache, incremental KV allocation, preemption, or UVA metadata.

  Averaging initial-run and replay results also does not quantify Prefix Cache effectiveness. Cache eviction and PD routing can prevent reuse, and per-phase cache-hit rates were not established.

  Both previously anomalous cases recovered in independent reruns. Execution-environment variability is therefore a plausible contributor, but concurrent testing has not been proven to be the cause.

  Larger samples, serial execution, and repeated measurements under consistent conditions are needed before treating these percentage differences as stable production expectations.


vllm_lt/pd/engine.py
```diff
@@ -0,0 +1,464 @@
...
+    def step(self):
+        if self.closed:
+            raise RuntimeError("PD engine is closed")
+        try:
+            self._poll()
+            if any(
+                time.monotonic() - w.created > self.config.request_timeout
+                for w in self.transfers.values()
+            ):
+                raise TimeoutError("PD request/transfer deadline exceeded")
```
hsliuustc0106, line 411: [P3] Any transfer exceeding `request_timeout` raises here, and `step()`'s handler then terminates **all** workers. Combined with the head-of-line break at line 356 (a request that can't be placed for `request_timeout/2` stops admission behind it), one oversized-but-accepted prompt — it passes the `add_request` capacity check against an idle peer, then starves at `_admit` under sustained load — fails every in-flight request after 300s. Fail-closed is defensible for memory safety (and the teardown ordering — stop writers before receivers — is correct), but request-scoped rejection (as `add_request` already does for oversized prompts) or at least documenting this blast radius in the flag help would make it operable.

vllm_lt/engine/preemption.py
```diff
@@ -0,0 +1,106 @@
...
+from vllm_lt.request import Stage
+
+
+class PreemptionManager:
+    def __init__(self, engine):
+        self.engine = engine
+        self.snapshots = {}
+        self.preemptions = self.resumptions = 0
+
+    def preempt(self, requester, *, priority_only=False):
```
hsliuustc0106, line 21: [P3] `preempt()` snapshots a RECURRENT victim without clearing `_signals[rid]`, so after resume, `_step_async` consumes the pre-preemption ticket's hazard score for the re-executed loop (llm_engine.py:390-404). I traced this to a safe outcome — the victim's loop is re-executed deterministically and the KV rewrite is idempotent, so old and new scores match — but it's a subtle invariant that would silently break if the recurrent pass ever became nondeterministic. The async preemption test uses the default `exit_threshold=1.0` (never populates `_signals`) and the sync test doesn't use the signal machinery at all, so a regression test with `async_scheduling + exit_threshold<1 + preemption` would pin this.

hsliuustc0106 (commented):
## Review (head `6bd9760`, base `main@8897fba`, merge base `488a038`)

**Verdict: no blocking correctness defects found.** Two non-blocking findings and one test-coverage gap below. The PD transfer protocol, KV ownership/lease lifecycle, incremental-allocation admission accounting, and the preemption path all traced clean against the base callers. Main has not touched any file this PR modifies since the merge base, so the cumulative diff against `main@8897fba` is identical to the stacked review.

### Design assessment

The motivating failure is real and measured: decode and prefill compete in one process, which this PR's own ITL/TTFT data shows (ITL P99 drops 43–70% when decode is isolated). The single-engine contract can't split GPU roles, so a new multiprocess coordinator is justified rather than a workaround. The chosen design — CPU coordinator holding exact capacity credits, workers as sole CUDA/NIXL owners, KV via registered-buffer WRITE with transfer leases deferring reclamation — has a coherent invariant story, and the code enforces it consistently; I traced cancel races at every phase transition (including cancel arriving before reserve) and the FIFO control-channel ordering holds. A one-process alternative (prioritized chunked prefill) would avoid the process lifecycle but not the interference. The honest weakness is bundling: seven features land together and the body correctly discloses that the benchmark does not attribute gains per feature — a reviewability cost, not a correctness defect.

### Findings

**[P2] Documentation asserts the opposite of what this PR implements.** The PR adds prefix caching, incremental allocation, preemption, and priority scheduling, but the docs at this head still declare them out of scope, and none of the new user-facing surface (`vllm-lt-pd-serve`, `--enable-prefix-caching`, `--incremental-kv`, `--kv-watermark`, `--scheduling-policy`, `--enable-preemption`, `--prefill-uva`) is documented outside `--help`:

- `README.md:21` — "preemption remain outside this version"
- `docs/design.md:67` — "Prefix sharing, eviction, swapping, and incremental reservation are outside this version."
- `docs/cdb_runtime.md:64` — "Incremental reservation/preemption and prefix sharing remain outside this change."

These files aren't touched by the diff, so they can't take inline comments — please update all three (and document the new entrypoint/flags) before merge.

**[P3] One stalled request terminates the entire PD engine** — inline comment on `vllm_lt/pd/engine.py`.

**[P3] Lossless preemption is untested on the async delayed-exit path it most affects** — inline comment on `vllm_lt/engine/preemption.py`.

### What I verified

- Full CPU suite at the frozen head: **238 passed, 11 skipped, 0 failures** (torch 2.8.0 CPU), including the 16 new CPU tests — lease-deferred reclamation, receive-commit gating, segment coverage, prefix/refcount/eviction, incremental-vs-reserved equivalence, sync preemption history, priority preemption.
- Traced core changes against every base caller: `_take` returning `None` is safe in `schedule()`; the PD prefill worker's direct `_take` call can't hit `None` because P allocations are always full-prompt; skipping `ensure_capacity` for CODA is correct (CODA samples, writes no KV); the `num_free_blocks`/watermark/reservation admission arithmetic is consistent and the no-preemption growth guarantee holds (live-request prefix publication doesn't shrink the evictable count); the `AsyncState` table re-upload is benign because block tables are append-only with immutable prefixes; the preemption snapshot/restore order (copy → release → free → requeue) is race-free under the pre-copy `synchronize()`.
- PD protocol: capacity credits stay exact across reserve/reject/cancel at every phase; slot/block accounting restores on all release paths; hidden-state and per-depth KV transfer addressing match on both ends, pinned by the fingerprint and `connect` geometry checks.

### Validation gaps

- **The CUDA/NIXL suite was not independently re-run for this review.** The four Triton-backend PD GPU tests were attempted on the evaluation host, but the GPU reservation window was lost to competing jobs (only one GPU free — below the 2-GPU test minimum, so the tests would skip). `flash_attn_4` is not installed on that host, so the FA4/UVA tests (`test_fa4_prefill_uva_prefix_growth_and_bank_reuse`, `test_prefill_uva_metadata_matches_reference_and_waits_for_consumer`) are unverifiable there regardless. The body's four-GPU runs are the author's evidence and are not pinned to this SHA.
- The benchmark section is a deployment-configuration comparison with candidly disclosed limitations (concurrent configs, averaged phases, no per-feature attribution). No speedup gate applies to a capability PR; the numbers are reported as configuration comparison, not qualification evidence — consistent with how the body frames them.

hsliuustc0106:
## Owner review (2nd — independent verification)

Reviewed `6bd9760be70e79ef41c5250b98f03a26d561a3da` against base `main@8897fba` (merge base `488a038`; full 2,863-line diff read plus frozen-tree verification; CPU-side checks executed locally; no GPU jobs run by this review). This confirms the earlier review's verdict — **no blocking correctness defects** — and adds two new P3s.

### Prior findings — all verified, still open

1. **[P2] Docs assert the opposite of the implementation** — re-confirmed at head: `README.md:21`, `docs/cdb_runtime.md:64`, `docs/design.md:67` still declare prefix caching / preemption / incremental reservation out of scope, and the new surface (`vllm-lt-pd-serve`, `--enable-prefix-caching`, `--incremental-kv`, `--kv-watermark`, `--scheduling-policy`, `--enable-preemption`, `--prefill-uva`) is documented only via `--help`.
2. **[P3] One stalled request terminates the entire PD engine** — re-confirmed by read: `PDEngine.step()` raises when any transfer exceeds `request_timeout` (including a request still waiting past the `request_timeout/2` head-of-line break in `_admit`), and the handler `_terminate()`s every worker.
3. **[P3] Preemption leaves `_signals` intact across snapshot/resume** — independently re-traced, agreeing with the safe-but-subtle conclusion: the retained hazard belongs to the already-executed round and is consumed exactly once after resume (`llm_engine` async signal consumption), so outputs stay correct; worth a comment or defensive clear to protect future refactors.

### New findings

**[P3] The `pd` package's private-coupling surface** (`vllm_lt/pd/worker.py`, `vllm_lt/worker/prefill_metadata.py`): workers bypass the engine API and drive internals directly — `scheduler.queues[Stage.WAITING].remove()`, a direct `scheduler._take(Stage.PREFILL)` call with manual queue surgery, `cache._get_allocation` / `cache._allocations`, the `pin_transfer(tid, tid)` self-lease convention, and `PrefillMetadataBank` constructing `_PreparedKVBatch` from privates. This extends the cross-module private reach-in pattern flagged in the #30 review to a much larger surface: any refactor of engine/scheduler/cache internals silently breaks PD. A narrow public seam on `LLMEngine`/`KVCacheManager` for worker-driven execution would contain the drift risk.

**[P3] Startup fingerprint streams the full model through host memory** (`vllm_lt/pd/worker.py`, the fingerprint digest loop): every parameter is copied via `.cpu().numpy().tobytes()` per worker — roughly two model-sized host copies (≈6 GB transient for Ouro-1.4B BF16) inside the startup window, multiplied by worker count. Hashing on-GPU into a digest, or hashing the checkpoint files, achieves the same P/D weights-match guarantee without the copies.

### Independently verified sound (full-diff read + tracing)

Coordinator capacity-credit exactness across reserve/reject/cancel at every phase; transfer-lease-deferred `free()` with completion via `unpin_transfer`; admission arithmetic (`required + reserved + headroom + cached_claim ≤ num_free_blocks`) preserving the no-preemption growth guarantee while incremental allocation is opt-in; `mark_imported_prefix` gating decode activation until all transfer sequences land; the P worker never samples (the final prefill chunk carries the hidden state through the registered slot and the request is never enqueued to CODA); prefix caching publishes only complete full-depth prompt blocks with all depths' pages together and keeps the final prompt token private (recomputed on hit against identical published KV — the whole-page-alias hazard from the LAST-EXITED contract is not reachable); scheduler hardening (`_take` requeue bounded by the original queue length, the `protected` set preventing self-preemption, CODA exempt from growth, RECEIVING-only states yielding instead of raising "no progress"); `AsyncState` table re-upload being append-only benign under incremental growth; preemption snapshot/restore ordering (page-at-a-time copies after `synchronize()`, then release → free → requeue).

### Validation (reviewed SHA `6bd9760`)

- Full CPU suite locally: **238 passed, 114 skipped, 0 failures** (torch 2.8.0 CPU; skips are NIXL / FA4 / GPU-gated / `lm_eval`, all unavailable on this host); Ruff lint + format clean.
- The CUDA/NIXL device suites and the four-GPU ShareGPT 1P3D/2P2D/3P1D campaign are author-reported and were not re-executed here. The benchmark section is honestly framed as a deployment-configuration comparison (concurrent configs, averaged phases, no per-feature attribution, anomalous cases rerun independently and disclosed); no speedup gate applies to a capability PR. Noted: the FA4/CUDA-graphs setup used in the campaign comes from #30 (now in base), so the measurements are reproducible from base+PR in principle, though they are not pinned to this SHA.

Issue #32, opened 2026-09-18
[RFC] Module boundaries and incremental refactoring of the RLT inference runtime


bjf-frz:
## Summary

This RFC proposes a module-by-module refactoring of vllm-lt, grounded in detailed code walkthroughs. The goal is to establish clear responsibilities, state ownership, and interfaces while preserving recurrent looped transformer (RLT) semantics: loop-level batching, exit policies, depth-aware KV caching, and prefill/decode (PD) disaggregation.

This is the overarching architecture and work breakdown proposal. It defines module boundaries, contracts, and refactoring dependencies before individual modules receive detailed reviews and implementation changes.

vLLM V1 provides an architectural reference for the scheduler/execution feedback loop and the separation of scheduling state from device execution. This does not imply adding a vLLM dependency, reproducing its full class hierarchy, or promising API/ABI compatibility.

## Motivation and scope

The implementation includes local inference, loop-level scheduling, asynchronous execution, multiple attention backends, cache management and preemption, and PD disaggregation. As these features have accumulated, several components have started accessing other components' internal queues, mutable request objects, and KV allocations.

The initial architectural review identified the following boundary concerns:

- Request objects combine logical progress with device tensors, RNG state, and asynchronous output placeholders. Multiple components mutate this state.
- The engine handles the execution loop, stage transitions, exit decisions, and output updates.
- The scheduler reads internal KV allocations, block tables, and reference counts to estimate resource costs.
- KVCacheManager combines logical block management, device storage, execution metadata, and attention dispatch.
- PD workers directly modify local scheduler queues and request stages and invoke private scheduling methods.
- Serving depends on internal scheduler/cache objects rather than a stable engine interface.

These observations identify architectural concerns, not necessarily confirmed correctness defects. Subsequent module reviews must establish concrete findings, their impact, and the evidence supporting each change.

The scope includes these capabilities as implemented or under development. Paths below identify responsibilities; some implementations may not yet be present on the default branch. Each module review must pin a reproducible commit or patch baseline before citing line numbers. This RFC does not claim that a complete line-by-line review has already been performed.

## Design principles

1. **Make state ownership explicit.** The scheduler owns scheduling-side request state and execution-result updates. EngineCore drives the scheduling/execution loop. Worker/ModelRunner owns device execution state.
2. **Separate logical resources from device operations.** KV allocation, references, and release conditions are distinct from tensor storage, copies, and attention execution.
[...]

PR #45, opened 2026-09-23
[REFACT] [M7 1/3] Extract the sampling module and pin request-local RNG behavior

ycjcx123:
## Problem and resulting behavior

Sampling arithmetic lived inline in `ModelRunner._sample_tensor` with no
unit tests pinning its contract, no independent documentation, and a dead
`_sample` wrapper. The RNG generator is stored on `Request.generator`
(scheduler-side), which disagrees with the RFC state table (worker-side),
but is functionally correct today because preemption retains the `Request`
object. This change extracts the algorithm into a testable `Sampler` class
and pins the current behavior with CPU contract tests, without touching RNG
ownership or the release path.

Refs #32. This is the first Sampling-focused M7 increment. RNG migration
and release-path split are M7 2/3; GPU validation is M7 3/3.

## Changes

- Extract `Sampler` class (`worker/sampler.py`) from
  `ModelRunner._sample_tensor`. The arithmetic is moved verbatim: greedy
  short circuit, FP32 logits promotion, top-k threshold mask, top-p
  cumulative-shift mask, and lazy generator creation. `Sampler` holds no
  per-request state; the caller passes the generator in and stores the
  possibly created one back.
- `ModelRunner._sample_tensor` becomes a thin delegate, preserving the
  existing `tests/test_async_pipeline.py` monkeypatch target and the async
  CODA path.
- Remove unused `ModelRunner._sample` (dead code, no callers).
- Add `tests/test_sampler.py`: greedy returns argmax and never creates a
  generator; top_k=1 matches greedy; top_k keeps tokens tied at the
  threshold; top_p retains at least one token; top_p=1.0 leaves full
  support; same seed reproduces the sequence; different seeds diverge;
  generator is reused and advanced, not reseeded.
- Add engine-level tests in `tests/test_engine.py`: a finished request
  drops its generator so the same ID with the same seed replays; greedy
  requests never create a generator at any step.
- Extend `tests/test_prefix_growth.py` with explicit RNG assertions:
  preemption preserves the generator object and state; resumption advances
  it without reseeding; termination releases it.
- Add `docs/sampling_walkthrough.md` following the RFC module structure
  (baseline, classes, functions, line-by-line coverage, findings, target
  design, validation). Link from `docs/README.md`.

## Behavior and compatibility

No behavior change. Sampling arithmetic, dtype promotion, kernel order,
RNG creation timing, and the absence of host synchronization are
unchanged. `SamplingParams` public fields and defaults are unchanged. The
serving protocol mapping is unchanged.

`ModelRunner._sample_tensor` remains as a delegate; its removal is planned
for M7 2/3 when the runner-side RNG registry lands and the test patches
`Sampler` directly.

GPU validation pending in M7 3/3.

## Validation

CPU evidence (macOS, Python 3.12, PyTorch 2.14.0):

| Suite | Result |
| --- | --- |
| `main @ 60af1cf` baseline (`test_engine.py test_prefix_growth.py test_async_pipeline.py`) | 47 passed, 15 skipped |
| Same three files after this increment (baseline 47 + 2 new engine tests) | 49 passed, 15 skipped |
| With `test_sampler.py` and `test_serving.py` added | 59 passed, 16 skipped |
| Full CPU suite (`tests/`) | 211 passed, 115 skipped |

Ruff: all checks passed; 3 files already formatted.

GPU tests skipped (`--run-gpu` not set). Device-side behavior under async
scheduling and CUDA streams will be covered by M7 3/3.


ycjcx123:
Noted that PR #44 (speculative decoding) landed worker/sampling.py with module-level functions (probabilities, generator_for, draw, sample_logits, rejection_sample) and a one-line delegate in _sample_tensor.
This PR keeps both modules for now: Sampler in worker/sampler.py as the M7 sampling boundary (stateless, generator passed in and out explicitly), and sampling.py unchanged for the speculative decoding path it was introduced with.

A follow-up PR will unify the two — either collapsing sampling.py's base functions into Sampler or vice versa — once the speculative decoding side is stable enough to retarget.

This branch has been rebased onto the latest `main` (`3314c1b`).

ycjcx123:
The previous rebase onto `3314c1b` resolved the `model_runner.py` conflict but missed updating the walkthrough to cover `worker/sampling.py` and the speculative decoding call sites. Two commit fixes that.

